PR #27, opened 2016-10-24
Added 'type' props to switch button

guillaumerxl:
Hi, 
as button tag is set to `type="submit"` by default, it causes problems when the switch is placed in a `<form>`.
I thought that passing `type` (if needed) to the `md-switch` could be useful?

Thanks for your work!

### DIFF
--- a/docs/src/pages/components/Switch.vue
+++ b/docs/src/pages/components/Switch.vue
@@ -36,6 +36,16 @@
           <md-switch v-model="checked7" id="my-test7" name="my-test7" class="md-primary" disabled>Light Blue Primary Color Disabled</md-switch>
         </div>
       </demo-example>
+
+      <demo-example label="Typed">
+        <form @click.stop.prevent="submit">
+          <md-switch v-model="checked8" id="my-test8" name="my-test8">Submit (default)</md-switch>
+        </form>
+
+        <div>
+          <md-switch type="button" v-model="checked9" id="my-test9" name="my-test9" class="md-primary">Button</md-switch>
+        </div>
+      </demo-example>
     </div>
 
     <div slot="code">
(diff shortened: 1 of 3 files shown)


marcosmoura:
Hi @guillaumerxl 

Thanks a lot to this PR. Yeah. You right! It is indeed a bug. The ```md-switch``` can have that problem inside forms. I think that the ```md-switch``` should always have the ```type="button"``` unless the user passes it as ```submit```. Also I've found in your PR that you forgot a ```alert``` there. I made some comments pointing those issues. If you agree to fix I can merge this directly to the develop branch, as the master branch is the current *release*.

Thanks again! :)

docs/src/pages/components/Switch.vue
```diff
@@ -36,6 +36,16 @@
           <md-switch v-model="checked7" id="my-test7" name="my-test7" class="md-primary" disabled>Light Blue Primary Color Disabled</md-switch>
         </div>
       </demo-example>
+
+      <demo-example label="Typed">
```
marcosmoura: Maybe this typed section can have the ```submit``` type instead of ```button```, as this is the default behaviour.

marcosmoura: The default behaviour will be ```type="button"``` and if the user change the type to ```submit``` in a switch inside a form this will submit the form. This will make things more explicit to a user that he needs to force the type to have a submit functionality, as the common scenario of a switch is to change a ```v-model``` value.

marcosmoura: What do you think?

docs/src/pages/components/Switch.vue
```diff
@@ -67,6 +77,28 @@
 &lt;/div&gt;
         </code-block>
       </demo-example>
+
+      <demo-example label="Typed">
```
marcosmoura: Same thing here.

src/components/mdSwitch/mdSwitch.vue
```diff
@@ -3,7 +3,7 @@
     <div class="md-switch-container" @click="toggleSwitch">
       <div class="md-switch-thumb" :style="styles" v-md-ink-ripple="disabled">
         <input type="checkbox" :name="name" :id="id" :disabled="disabled" :value="value">
-        <button class="md-switch-holder"></button>
+        <button :type="type" class="md-switch-holder"></button>
```
marcosmoura: If the user don't pass the type we must assume that this will have ```type="button"```
We need to change that to something like that:
``` html
<button :type="type || 'button'" class="md-switch-holder"></button>
```

src/components/mdSwitch/mdSwitch.vue
```diff
@@ -24,7 +24,8 @@
       name: String,
       value: Boolean,
       id: String,
-      disabled: Boolean
+      disabled: Boolean,
+      type: String
```
marcosmoura: Or you can pass a default property right here on the ```type``` prop.

marcosmoura:
Hi @guillaumerxl 

Thanks a lot to this PR. Yeah. You right! It is indeed a bug. The ```md-switch``` can have that problem inside forms. I think that the ```md-switch``` should always have the ```type="button"``` unless the user passes it as ```submit```. I made some comments pointing those issues. If you agree to fix I can merge this directly to the develop branch, as the master branch is the current *release*.

Thanks again! :)

guillaumerxl:
Yes, I think it's better to have the `type="button"` behaviour by default too, as, like you said, a switch is not aiming to submit a form. I don't see use cases where a switch should submit a form but I think that having the possibility to do it is better.

marcosmoura:
Yes. Exactly my point! Cool! Can you fix those things? After that I will update the changelog for the next *fix* version.

guillaumerxl:
Yes sorry I answered last night just before going to bed, I'm on it :)

marcosmoura:
Oh!! Great man! Approved! I will add this in the changelog later.

Thanks a lot!

PR #28, opened 2016-10-24
Add babel-cli as dev dependecy

lucassouza1:
We need to be sure babel-cli is installed as it is required for dev.


marcosmoura:
Oh yeah. You're right. I will change the branch to ```develop``` to point to the next minor version.
Thanks a lot!

Issue #40, opened 2016-11-01
md-input not reactive when initial value empty


deulos:
The md-inputs do not acquire the value of the v-model if their initial value is empty. Example:

In template:
```
...
<input v-model="foo">

<md-input-container>
  <label>Label</label>
  <md-input type="text" v-model="foo" required></md-input>
</md-input-container>
...
```

In code:
```
...
data() {
  return {
    foo: null
  };
},

created() {
  setTimeout(() => {
    this.foo = 'bar'
  }, 2000);
}
...
```

The simple input will update it's value after 2 secs but the md-input will remain empty.

This does not happen when foo has any value with length (not null nor empty string).

marcosmoura:
Oh. Quite strange. I'll take a deep look at this and I will come back with a fix. ASAP. 

Thank you for sending this.

deulos:
No problem, I gues is due to not watching the value, so the class taking the label above is not applied

damienix:
I have the same problem with combination with vuex. It looks like md-input is never updated when you update value programmatically using a model.

marcosmoura:
Already implemented and will be shipped on the very next release (v0.3.0)

Issue #41, opened 2016-11-01
md-button ignores type prop


deulos:
The following md-button
`<md-button type="submit" class="md-raised md-primary">Save</md-button>`

Is rendered as
`<button type="button" class="md-button md-raised md-primary">Save</button>`

So the type is always rendered as "button"

marcosmoura:
Hey. You're right. This is a ugly bug. Thanks for the PR and for reporting this.

PR #42, opened 2016-11-01
Fixed the issues #40 and #41

deulos:
This will fix the issue [`#40`](https://github.com/marcosmoura/vue-material/issues/40)
This will fix the issue [`#41`](https://github.com/marcosmoura/vue-material/issues/41)

### DIFF
--- a/src/components/mdButton/mdButton.vue
+++ b/src/components/mdButton/mdButton.vue
@@ -14,7 +14,7 @@
       let options = {
         staticClass: 'md-button',
         attrs: {
-          type: hasLink || 'button',
+          type: hasLink || this.type || 'button',
           disabled: isDisabled
         },
         directives: [{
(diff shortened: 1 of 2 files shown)


src/components/mdButton/mdButton.vue
```diff
@@ -14,7 +14,7 @@
       let options = {
         staticClass: 'md-button',
         attrs: {
-          type: hasLink || 'button',
+          type: hasLink || this.type || 'button',
```
marcosmoura: The correct behaviour of this has to be:
```type: this.type || 'button',```
The ```hasLink``` thing is unnecessary.

marcosmoura: I merged another PR that fixed this issue. So, after you correct this small thing, I can merge this one.

marcosmoura:
I will change the way that input and input container handle classes. I will close this PR because those changes related to methods changing classes directly will become obsolete. Thank you!

PR #50, opened 2016-11-09
pass on  DOM event for checkbox and radio

kristianmandrup:
Please do the same for other clickable/changeable components. Thanks :)


kristianmandrup:
This works:

`<input type="checkbox" @change="onChange($event)" v-model="myVal"/>`

```
  methods: {
    onChange (event) {
      console.log(event)
    },
```

For `@click` used currently it passes a `MouseEvent` which is not very useful however. Also click would not work in case you use keyboard or other non-input devices or even DOM manipulation.
Please change event handler to `@change` instead. Cheers!

marcosmoura:
That's good. I will change and fire some other events when a value change too. I will fix those things on the 0.4.0 version. Thank you!

PR #52, opened 2016-11-10
remove dotted outline around button in firefox

yeknava:


marcosmoura:
That's nice. O completely forgot about this rule. Thank you!

Issue #88, opened 2016-11-17
Issue with md-table-pagination on an Electron app


pauloramires:
Hi,
while using the md-table-pagination component on an electron app I got the following error:
<img width="540" alt="electron-table-issue" src="https://cloud.githubusercontent.com/assets/1428629/20382744/563ef966-ac94-11e6-9f76-d5c782fdcf7c.png">
I have identified that this error is thrown by src/utils/getClosestVueParent.js, and that if the $parent argument is initialised as an empty object then it goes away, and the component works perfectly.
This is working for me:
```javascript
let getClosestVueParent = ($parent = {}, cssClass) => { ... }
```
I am using [electron-vue](https://github.com/SimulatedGREG/electron-vue) by **@SimulatedGREG**.
Could you have a look at it? 
PS: it would be useful if you could also include a non-uglyfied version of vue-material.js in the module package (e.g. vue-material.debug.js) so that we could more easily contribute to the resolution of this kind of issues without having to compile the module from source.

Many thanks!
Paulo

javaday:
I just ran into this myself.

javaday:
After seeing the fix suggested by @pauloramires, I figured out that the root Vue component has an undefined $parent property. So I added "this.$parent = {};" to the root component's mounted handler and I don't get the error anymore.

So, this seems to be a viable work around until the Paulo's pull request is merged.

marcosmoura:
Oh yeah. I will take a look at this. This will be fixed on the very next fix version.

pauloramires:
Hi Marcos,
after updating to v0.3.3 I still get the 'Cannot read property '$el' of undefined' error.
I noticed that the culprit is the **md-table-pagination** component. If I remove it from the view, then the error is gone.

PR #90, opened 2016-11-17
Add material.debug.js to dist folder

pauloramires:
Hi, here is a spontaneous contribution for issue #88.
Feel free to use it.

Regards,
Paulo


marcosmoura:
Hi @pauloramires. Thank you so much for this PR. I will decline for two reasons:
- The fix related to the $parent thing is kind of wrong. You did a good thing, but the getClosestVueParent should stop when reach the $root. We should check that and assume that this is not going to be a problem. The md-table should work with or without the card and right it's not. :)
- I liked the idea of the debug file. Can you please remove the fix and leave only the debug part? After that I will merge this instantly!

src/core/utils/getClosestVueParent.js
```diff
@@ -1,4 +1,4 @@
-let getClosestVueParent = ($parent, cssClass) => {
+let getClosestVueParent = ($parent = {}, cssClass) => {
```
marcosmoura: Can you revert this part?

pauloramires:
Done

PR #109, opened 2016-11-23
Add extra "align" directions that do not cover menu trigger.

damienix:
I needed a menu list that do not cover menu trigger. You can see how it should look at https://material.google.com/components/menus.html. I didn't bother to create issue for that.

components_menus_usage3do

I think the code is quite obvious but I have one concern about the API. I was wondering it it would be better to just add another direction constant strings (which I did) or maybe as another prop called "align" for this purpose. Also if you have better name for this feature I'm open for the change.

### DIFF
--- a/src/components/mdMenu/mdMenu.vue
+++ b/src/components/mdMenu/mdMenu.vue
@@ -22,6 +22,18 @@
         type: String,
         default: 'bottom right'
       },
+      mdAlignTrigger: {
+        type: Boolean,
+        default: false
+      },
+      mdOffsetX: {
+        type: [Number, String],
+        default: 0
+      },
+      mdOffsetY: {
+        type: [Number, String],
+        default: 0
+      },
       mdCloseOnSelect: {
         type: Boolean,
         default: true
@@ -69,9 +81,11 @@
         this.menuContent.classList.add('md-size-' + size);
       },
       addNewDirectionMenuContentClass(direction) {
-        this.menuContent.classList.add('md-direction-' + direction.replace(/ /g, '-'));
+        if (!this.mdAlignTrigger) {
+          this.menuContent.classList.add('md-direction-' + direction.replace(/ /g, '-'));
+        }
       },
-      getPosition(vertical, horizontal, align) {
+      getPosition(vertical, horizontal) {
         let menuTriggerRect = this.menuTrigger.getBoundingClientRect();
 
         let top = vertical === 'top'
@@ -82,7 +96,10 @@
           ? menuTriggerRect.left - this.menuContent.offsetWidth + menuTriggerRect.width
           : menuTriggerRect.left;
 
-        if (align) {
+        top += parseInt(this.mdOffsetY, 10);
+        left += parseInt(this.mdOffsetX, 10);
+
+        if (this.mdAlignTrigger) {
           if (vertical === 'top') {
             top -= menuTriggerRect.height;
           } else {
@@ -100,28 +117,12 @@
             position = this.getPosition('bottom', 'left');
             break;
 
-          case 'bottom left align':
-            position = this.getPosition('bottom', 'left', true);
-            break;
-
-          case 'bottom right align':
-            position = this.getPosition('bottom', 'right', true);
-            break;
-
           case 'top left':
             position = this.getPosition('top', 'left');
             break;
 
-          case 'top left align':
-            position = this.getPosition('top', 'left', true);
-            break;
-
           case 'top right':
             position = this.getPosition('top', 'right');
-            break;
-
-          case 'top right align':
-            position = this.getPosition('top', 'right', 'align');
             break;
 
           default:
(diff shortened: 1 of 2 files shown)


src/components/mdMenu/mdMenu.vue
```diff
@@ -63,81 +63,73 @@
...
       },
       addNewSizeMenuContentClass(size) {
         this.menuContent.classList.add('md-size-' + size);
       },
       addNewDirectionMenuContentClass(direction) {
-        this.menuContent.classList.add('md-direction-' + direction.replace(' ', '-'));
+        this.menuContent.classList.add('md-direction-' + direction.replace(/ /g, '-'));
       },
-      getBottomRightPos() {
+      getPosition(vertical, horizontal, align) {
```
damienix: I merged all 4 functions into 1 to avoid dupplication. I think it looks clean now.

marcosmoura:
Man. Sorry about the delay to answer this, but I must say: You just nailed it. Liked a lot the 'dedupe' thing. 

I think that another property (with true value by default) will be more easy to understand by other users. Something like:
```
    <md-menu md-direction="bottom left" :md-parent-align="false">
      <md-button md-menu-trigger>Open</md-button>

      <md-menu-content>...</md-menu-content>
    </md-menu>
```

Or even with:
```
    <md-menu md-direction="bottom left" md-offset="0 48px">
      <md-button md-menu-trigger>Open</md-button>

      <md-menu-content>...</md-menu-content>
    </md-menu>
```

Or maybe both. I think this will make the component more flexible.
What do you think? Do you want to contribute with this?

damienix:
No problem. :md-parent-align looks good to me, I'll change this. Just mind that making true a default value will break the compatibility, but if you think it's the more common use case I can do this (it's still 0.X anyway).

md-offest might actually make sense as well. I'm not sure about the syntax though as parsing expression as in the example might be error prone. How about sth like this:
```
    <md-menu md-direction="bottom left" :md-offset-y="48">
      <md-button md-menu-trigger>Open</md-button>

      <md-menu-content>...</md-menu-content>
    </md-menu>
```
What do you think?

marcosmoura:
Oh. I said it wrong. Make it FALSE is better. 

``` vue
    <md-menu md-align-trigger> <!-- I think that trigger will be better than parent -->
      <md-button md-menu-trigger>Open</md-button>

      <md-menu-content>...</md-menu-content>
    </md-menu>

    <md-menu md-direction="bottom left" :md-offset-x="48px">
      <md-button md-menu-trigger>Open</md-button>

      <md-menu-content>...</md-menu-content>
    </md-menu>
```

Oh yeah. Make more sense to make it separately. I liked it!
What do you think?
:md-offset-y and :md-offset-x accepting only 'px' or the number without the unit. (I saw some users trying to pass 'dp' in whiteframe units)
Like this: 
``` vue
    <md-menu md-direction="bottom left" :md-offset-y="48">
      <md-button md-menu-trigger>Open</md-button>

      <md-menu-content>...</md-menu-content>
    </md-menu>

    <md-menu md-direction="bottom left" :md-offset-x="48px">
      <md-button md-menu-trigger>Open</md-button>

      <md-menu-content>...</md-menu-content>
    </md-menu>
```

I think that this will be a GREAT improvement over menus.
What do you think? Sounds reasonable? :D

damienix:
I like `md-align-trigger`. It's just enough to handle boolean and looks very neat. Name also makes more sense.

I'm not sure about allowing 2 ways of inputting offet valus. I think it's better not to let the user think "should I pass 10 or 10px there?". If we just make it a number and give proper example then I think they should understand. They will be copypasting the example anyway ;)

damienix:
Ok, see the commit message for the summary. I partially followed your advice and allowed to use number-string in md-offest attributes to allow simpler syntax without semicolon.

Also I didn't bother to use more inteligent parsing than parseInt which has some side effects like '100vh' value understood as '100px'. Let me know if you think that better parsing is needed.

P.S. I'll update the example code once we finish the library code. Suggestions for better examples are welcome.

marcosmoura:
Yeah. You're right. We only need to explain on the API. And this should be with v-bind or without it. Sounds good for me.

Do you think that you can develop those changes?

------------

Also I wanna give you a small preview of the 'new' documentation website. I think this will be a deal breaker for this project.
New Docs

Do you have some suggestions related to the API of menus?

marcosmoura:
Oh. You did it quickly! Do you have some other improvements over what you already developed or I can do the Code Review?

damienix:
I think I'm done with my changes. I've updated examples code. Do you need anything more to merge this?

API for Menus looks good to me as for the features you have.

With regards to new documentation, I really like it. Switching between "Examples" and "Code" tabs is currently quite painfull. In the new version looks very simple. Also it's slightly densier and clear API reference is also useful, I already missed some API options in the past (it's hard to find them all in the examples), now it's all clear. Good job!

marcosmoura:
Merged!

damienix:
Great thanks!

marcosmoura:
Thank you for this PR. I liked this discussion a lot. I hope that you can contribute more. :)

damienix:
I liked cooperation with you as well. I think there is a chance that I will contribute, but we'll see ;)
Thanks.

Issue #114, opened 2016-11-25
error message if md-sort-by is not set


Cabbaggio:
Error Message: Uncaught (in promise) TypeError: Cannot read property 'removeEventListener' of undefined

http://codepen.io/anon/pen/mOMbwB

marcosmoura:
Oh. That's bad. I think the ```<md-table-head>``` is having a problem with the ```md-ink-ripple``` component. I will take a deep look. Sorry about that.

PR #118, opened 2016-11-26
unregisterMouseEvent accepts the element as parameter

Cabbaggio:
The mdInkRipple-directive passes the elemnt as an param. But the param option is missing.
https://github.com/marcosmoura/vue-material/blob/master/src/core/components/mdInkRipple/index.js#L114

It's fix for #114 not #109 🙈


marcosmoura:
Oh. Thanks for the fix. The ripple directive and component will be improved on the nexts releases.

PR #155, opened 2016-12-05
[update] Backdrop component classes customization

jtouzy:
Today we have a "shadowed" backdrop : This is the default backdrop. 
But sometimes we need to have a fullscreen "colored" backdrop in some cases (examples of Login Dialog in Fullscreen).

Example : 

capture d ecran 2016-12-06 a 00 00 51

Demo can be found here: http://demo.geekslabs.com/materialize/v2.1/layout03/ui-buttons.html#! (menu Users/Login)

So i add the ability to add custom classes to the mdBackdrop component without touching the default class `md-backdrop`.
You can do `this.$refs.dialog.$refs.backdrop`... Not really clean. I think the better way is to send classes object directly to the dialog we need.

What do you think?

### DIFF
--- a/src/components/mdDialog/mdDialog.vue
+++ b/src/components/mdDialog/mdDialog.vue
@@ -4,7 +4,7 @@
       <slot></slot>
     </div>
 
-    <md-backdrop class="md-dialog-backdrop" :class="classes" v-if="mdBackdrop" ref="backdrop" @close="mdClickOutsideToClose && close()"></md-backdrop>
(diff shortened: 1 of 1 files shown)


jtouzy:
After thinking, i've found another simple solution : Just put a class on the dialog, and then customize the `.my-class .md-backdrop` class... Sorry for the inconvenience :-D

marcosmoura:
Oh. Yeah. Don't worry. I really like that template layout. Are you implementing something like that with Vue Material? Let me know!

jtouzy:
@marcosmoura I'll send you a mail :-)

PR #166, opened 2016-12-06
Fix table pagination

### LINKED ISSUE #163: mdTablePagination - No way to know if next page or prev page?
mdTablePagination emits `pagination` and `page` events, but there's no way to know if we should move to the next page or the previous page (unless I'm missing something?).

Here's the code responsible for that:
https://github.com/marcosmoura/vue-material/blob/master/src/components/mdTable/mdTablePagination.vue#L73

The same method is called regardless of if it's the next or the previous button that has been pressed.

This could be fixed by passing an argument to that method [here](https://github.com/marcosmoura/vue-material/blob/master/src/components/mdTable/mdTablePagination.vue#L15) and [here](https://github.com/marcosmoura/vue-material/blob/master/src/components/mdTable/mdTablePagination.vue#L11).

I can create a pull request to implement this if that's okay?

jonataswalker:
This PR fixes #163.

### DIFF
--- a/src/components/mdTable/mdTablePagination.vue
+++ b/src/components/mdTable/mdTablePagination.vue
@@ -6,13 +6,13 @@
       <md-option v-for="amount in mdPageOptions" :value="amount">{{ amount }}</md-option>
     </md-select>
 
-    <span>{{ (currentSize - currentSize + 1) * currentPage }}-{{ currentSize }} {{ mdSeparator }} {{ mdTotal }}</span>
+    <span>{{ ((currentPage - 1) * currentSize) + 1 }}-{{ subTotal }} {{ mdSeparator }} {{ mdTotal }}</span>
 
-    <md-button class="md-icon-button md-table-pagination-previous" @click="changePage" :disabled="currentPage === 1">
+    <md-button class="md-icon-button md-table-pagination-previous" @click="previousPage" :disabled="currentPage === 1">
       <md-icon>keyboard_arrow_left</md-icon>
     </md-button>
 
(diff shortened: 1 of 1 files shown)


src/components/mdTable/mdTablePagination.vue
```diff
@@ -6,13 +6,13 @@
...
-    <span>{{ (currentSize - currentSize + 1) * currentPage }}-{{ currentSize }} {{ mdSeparator }} {{ mdTotal }}</span>
+    <span>{{ ((currentPage - 1) * currentSize) + 1 }}-{{ subTotal }} {{ mdSeparator }} {{ mdTotal }}</span>
 
-    <md-button class="md-icon-button md-table-pagination-previous" @click="changePage" :disabled="currentPage === 1">
+    <md-button class="md-icon-button md-table-pagination-previous" @click="previousPage" :disabled="currentPage === 1">
       <md-icon>keyboard_arrow_left</md-icon>
     </md-button>
 
-    <md-button class="md-icon-button md-table-pagination-next" @click="changePage" :disabled="currentSize * currentPage >= totalItems">
+    <md-button class="md-icon-button md-table-pagination-next" @click="nextPage" :disabled="currentSize * currentPage >= mdTotal">
```
marcosmoura: The disabled rule should use the `totalItems` variable. If the user don't pass the total the pagination must assume that this is "infinite". Some cloud based database don't support the "total" items to be retrieved in it's queries if there is a huge amount of data on the server. I changed this for you here!

Issue #165, opened 2016-12-06
Internet Explorer issue with compatibility


Sarpedion:
Greetings,

I've been testing if my application works equally well on all browsers, and I've noticed the following issue on Internet Explorer 11. Google Chrome and Mozilla Firefox work extremely well and have no compatibility issues that I've noticed. Furthermore, I've checked on two different machines, and both had same issue.

The issue can be replicated by going to [tabs demo page](https://vuematerial.github.io/#/components/tabs) on IE 11, where tabs component is not rendered:

![image](https://cloud.githubusercontent.com/assets/16826299/20939074/b45c3d8e-bbed-11e6-923e-e3af6021e70b.png)

And that is likely related to the errors that are shown in console:
![image](https://cloud.githubusercontent.com/assets/16826299/20939103/ce3fbeba-bbed-11e6-9b98-09efb5b48b9c.png)

I know, IE is generally bad, but there might actually be someone that wants to use it.

Thanks!

marcosmoura:
Wow. That's bad. IE 11 is bad but is a Vue Material target. I will take a look at this. Thank you for reporting that!

vuchl:
@Sarpedion can you try the demo page with the [`polyfill`](https://qa.polyfill.io/v2/docs/) from here?

marcosmoura:
The problem is with the `Object.values` functions. I will replace it with another method. :)

vuchl:
well it's allready a polyfill. maybe looking for an alternative rather then implementing your self is a safe bet.

Sarpedion:
@vuchl I must be honest, I do not know how to use polyfill. I've just included it as a script on index.html, and I haven't noticed any improvement. Thanks for the advice anyways, I am going to read docs a bit on how to implement it. :)

vuchl:
@Sarpedion that should be enough I believe.

PR #167, opened 2016-12-07
[fix] md-tabs in IE/Safari + Fix tab @change event emitted

jtouzy:
Fix issue #165 by using for...in instead of Object.values()
This bug also appears in Safari (original Issue only on IE).


jtouzy:
Add the md-tabs $emit() fix #168

marcosmoura:
Just to be clear. The Object.values is cleaner than a for .. in. I will enable the babel-2017 preset for future features. Thank you, again!

PR #173, opened 2016-12-08
Support for theming on Cards

mateusmcg:
Added a simple change to be able to add theme on a entire card or in each internal component independently


mateusmcg:
Sorry about that. I thought it could be good to have that kind of flexibility but i guess you're right, that can lead to some problems.

Now theming is only possible on the `<md-card>` tag

PR #175, opened 2016-12-08
Adjust menu offset position when page scrolling occurs

aagahi:
Menu position could appear on an incorrect position when page is scrolled as top/left are absolute position they should take pageX/Y offset position into account.


marcosmoura:
I've merged but I changed the base. Should always be against the develop branch.

aagahi:
Ok, np problem.

PR #178, opened 2016-12-09
add Spinner component (based on materialize element)

throrin19:
Add the spinnel indeterminate element like materializecss component : http://materializecss.com/preloader.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .DS_Store
 node_modules/
 npm-debug.log
+
+.idea/
--- a/docs/src/App.vue
+++ b/docs/src/App.vue
@@ -102,6 +102,10 @@
                 </md-list-item>
 
                 <md-list-item class="md-inset">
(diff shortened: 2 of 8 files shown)


marcosmoura:
Hi @throrin19.

Thank you for your PR. Unfortunately I will have to decline this PR. The idea of this library is to develop everything from scratch and code from other packages will not be integrated here. If you want to develop the spinner using your own code I will be glad to merge everything. I will update the PR guideline with this information. I'm sorry.

throrin19:
But after research, the code is just css and html. I kept only the basic functionnal design and this is the only code compatible with ie10+. Otherwise, to create this spinner, the only solution is to use a `svg` element and, after test, it's not a good idea

marcosmoura:
Hey @throrin19. The spinner component is already implemented and will be available on the very next version. But the progress bar is still on the backlog. Do you think that you could implement this? Thanks.

PR #180, opened 2016-12-11
[fix] Handling md-list-item click

jtouzy:
Hi,

In MdListItem, we cannot handle the @click native-event because of the declaration of the listItemSpec. 
And then, we cannot do ``@click.self`` or others because the native event is always ``undefined``, this is the vue.js code generated when we add ``@click.self`` :

```javascript
on: {
   "click": function($event) {
      if ($event.target !== $event.currentTarget) { return; }
         _vm.navigateToDetail(item)
   }
}
```

The PR just send the $event on click event in items.

### DIFF
--- a/src/components/mdList/mdListItem.vue
+++ b/src/components/mdList/mdListItem.vue
@@ -16,8 +16,8 @@
       let listItemSpec = {
         staticClass: 'md-list-item',
         on: {
-          click: () => {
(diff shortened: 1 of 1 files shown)


marcosmoura:
Hi @jtouzy 

I've missed this! Thanks for the PR!

PR #190, opened 2016-12-13
fix : tablePagination bug

JasonOneYuan:
to display the right number , not always Numer.MAX_SAFE_INTEGER or
“false”

### DIFF
--- a/src/components/mdTable/mdTablePagination.vue
+++ b/src/components/mdTable/mdTablePagination.vue
@@ -48,7 +48,7 @@
         subTotal: 0,
         currentSize: parseInt(this.mdSize, 10),
         currentPage: parseInt(this.mdPage, 10),
-        totalItems: !isNaN(this.mdTotal) && Number.MAX_SAFE_INTEGER
+        totalItems: isNaN(this.mdTotal)?Number.MAX_SAFE_INTEGER:parseInt(this.mdTotal, 10)
       };
     },
     computed: {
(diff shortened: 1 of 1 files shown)


src/components/mdTable/mdTablePagination.vue
```diff
@@ -48,7 +48,7 @@
         subTotal: 0,
         currentSize: parseInt(this.mdSize, 10),
         currentPage: parseInt(this.mdPage, 10),
-        totalItems: !isNaN(this.mdTotal) && Number.MAX_SAFE_INTEGER
+        totalItems: isNaN(this.mdTotal)?Number.MAX_SAFE_INTEGER:parseInt(this.mdTotal, 10)
```
marcosmoura: The user can still pass a String to the `mdTotal`. The `totalItems` should be used only to disable the 'next' button. The fix is inside the line 9. Just change this `totalItems` to `mdTotal` on the template.

JasonOneYuan: I see , but i think this change is still necessary , because 
if the user pass a `NAN String` to `mdTotal `, the `totalItems` will return `false` and the `disable` condition in line 15 (e.g 10*5 >= false) will always be `true` .  
if the user pass a `numberlike String` to `mdTotal` , the `totalItems` will return `MAX number` and the `disable` condition will always be `false`

marcosmoura: I agree with you! I said it wrong. What I meant was the fix is not the awesome part that you did, but the template too.
Could you fix the template part?

JasonOneYuan: offcause

marcosmoura: Great. Thanks!

PR #211, opened 2016-12-16
lazy update ink's size

vanhoavn:
Sometime the ink's size cant be populated when document is loading, so we need to calculate the size later.


marcosmoura:
Hello @vanhoavn.

Thank you for the PR. I will deny it for one reason: **I'm developing** a new ripple system.
The idea is to have a reactive ripple, with better touch response, better perfomance, component based and optional in every single component, with a flag to disable it.
I'm sorry about that.

marcosmoura:
Ah. I forgot to mention. Will be present in the very next version.

PR #252, opened 2016-12-25
Fix mdTheme.vue filename case

korylprince:
Building was failing because `src/core/components/mdTheme/index.js` was trying to import `MdTheme` instead of `mdTheme`. My guess is you're on a case-insensitive filesystem, but for us linux folk it matters :)

If you want to just do this in your own commit, that's cool.


marcosmoura:
Oops. My bad!
Thank you for fixing it!

PR #270, opened 2016-12-29
check if .md-ripple exists (it doesn't if not using md-button)

korylprince:
The title pretty much says it all, but `this.snackbarElement.querySelector('.md-ripple')` is null when an `md-button` is not included causing an error in the console.

### DIFF
--- a/src/components/mdSnackbar/mdSnackbar.vue
+++ b/src/components/mdSnackbar/mdSnackbar.vue
@@ -67,12 +67,14 @@
       }
     },
     methods: {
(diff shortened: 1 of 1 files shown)


korylprince:
Sorry about the the indentation. I have mine set to 4 because I'm often writing python. Also wasn't sure about whether to create a cache variable or not. 

If I do another pull request I'll keep these in mind :)

Thanks for being responsive! You've done an incredible amount of work on this project and I've found it really useful.

Issue #277, opened 2016-12-31
input label not reset when set v-model variable empty


nothinggift:
`<md-input-container>
    <label>Initial value</label>
    <md-input v-model="initialValue"></md-input>
  </md-input-container>`

Input something and set initialValue empty,  the label will not reset , test on 0.5.2 .

korylprince:
I've seen this problem as well. I've created a codepen showing this issue.

Type some content into the input, then click the "Break!" button to set the `v-model` to `null`. The label doesn't move down to fill the input like it would if you deleted all the text manually.

korylprince:
@nothinggift, I submitted a PR to fix this issue, so as long as it looks good to @marcosmoura, this should appear in the next release. I'm also realizing I never actually pasted the codepen in my previous post. For reference, it is: http://codepen.io/anon/pen/XpraNV

nothinggift:
It's useful,thanks!

marcosmoura:
PR merged!

pluschen:
still present in version 0.7.1

PR #281, opened 2017-01-01
fixes md-input-container label bug #277

korylprince:
It appears the watcher for `value` (in `common.js`) runs before the input's value has synced through the v-model. Thus when the watcher for `value` runs, the value provided to the function is the new value, and `this.$el.value` is still the old value.

In the case where the input value is non-`null` and you set the value to `null`, the function:

```
    setParentValue(value) {
      this.parentContainer.setValue(value || this.$el.value);
    },
```
Actually sets the parent value to the old value (since `value` is `null` and `this.$el.value` is still the old non-`null` value.) This means the parent sees the old value, it's `hasValue` is computed as true, and the `md-has-value` class is still applied, leaving the label above the input instead of in it as a placeholder.

This PR just explicitly sets `this.$el.value = value` before calling `setParentValue` in the `value` watcher, fixing the issue.

### DIFF
--- a/src/components/mdInputContainer/common.js
+++ b/src/components/mdInputContainer/common.js
@@ -8,6 +8,7 @@ export default {
   },
   watch: {
     value(value) {
+      this.$el.value = value;
       this.setParentValue(value);
     },
     disabled() {


marcosmoura:
Sorry about the delay to answer this. The new years party here in Brazil is craaazy!
Thanks a lot for this contribution. I'm approving!

PR #287, opened 2017-01-04
support for font awesome and many others css icon themes

sombriks:
Hi,

there are many icon sets over there, many of them have zero or incomplete ligature support.

Yet there are plenty of them around, and i think that vue-material could benefit from even basic support for them.

This pull request intends to start to offer support for those 3rd party icon sets, so i'l up to the developer to either stick with material icons or to consume many other icon sets available on the internet.

Please review this pull request, i'll promptly try to address any change it needs.

For your time thank you.


src/components/mdIconCSS/mdIconCSS.vue
```diff
@@ -0,0 +1,18 @@
...
+</template>
+
+<style lang="scss" src="./mdIconCSS.scss"></style>
+
+<script>
+  import theme from '../../core/components/mdTheme/mixin';
+
+  export default {
+    props: {
+      cls: String
```
Morgul: You actually don't need to use `cls` as a property. According to [the docs](https://vuejs.org/v2/guide/class-and-style.html#With-Components) classes put on a component get combined and rendered correctly. So, this:

```html
<md-icon-css class="md-primary fa fa-fw fa-home"></md-icon-css>
```

will get rendered as:

```html
<i class="md-icon-css md-primary fa fa-fw fa-home"></i>
```

sombriks:
Hi @Morgul thanks for this review.

I agree with you, an entire new component is overkill. i need help however on define an idiom on how to inform md-icon when to use material icon css or user defined icon css. 

I'll rethink this patch and the last dangling bit is this case: how to tell if we should use standard material icon css or if we should use developer-defined icon css

marcosmoura:
Hi @sombriks 

I don't see this as a new component, like @Morgul said.
I think that fonts without ligatures are always bad. You will have an empty element with no fallback. The ligatures enable a way to have better content for screen readers too.
Take a look at this example: https://gist.github.com/jareware/76beb201c717fd6333c7

In the case of non existent icons, you can use this library too: https://materialdesignicons.com/

You can also use Font Awesome using ligatures, converting here on this website: https://icomoon.io/

But notice that Font Awesome do not follow the Material Design specs for icons, so you'll end up with an inconsistence here. :)

Thank you!

Morgul:
@marcosmoura Actually, materialdesignicons.com support would be amazing to have. Just a brief glance, and it has support for several of the missing icons I wanted. Would there be any work associated with making `md-icon` work with it? Perhaps an example/explanation for use added in the icons section?

sombriks:
Hi @marcosmoura !

Indeed legacy icon fonts does not fulfill material design specs. On the long run, most icon sets hopefully will support it correctly.

In that case i'll bake a brand new patch, just for the docs section, explaining how to convert legacy icons into modern ones. Also i'll try to hit develop instead master ;-)

On the other hand @Morgul pointed a quite good icon set and he can correct me on this, but mdi lacks ligature support too.

One of my biggest concern is about how we could make such change on md-icon without cause too much noise on native material icon support and at the same time not let 3rd party css iconset gets somehow modified by material icon default css.

Before i make any other pull i would like to hear you guys about this. :thumbsup:

marcosmoura:
Well. I think that Angular Material handle well the external icons. I don't think that needs to much effort to do the integration without changing much the core component. Maybe you can create a optional prop for the iconset name inside the md-icon component:

This property default value will be `material-icons`, and if the user passes another name, like `fa`, it will render the correct icon:

``` vue
<md-icon md-iconset="fa" class="fa-home"></md-icon>
```

That will become:

```
<i class="md-icon fa fa-home"></i>
```

I don't think that we will need to disable the ligatures support.

BTW Thankyou @sombriks and @Morgul for raising this discussion! o/

sombriks:
@marcosmoura amazing!

Please wait for the next PR :-)

Morgul:
@marcosmoura You're welcome.

A suggestion for the proposed syntax:

```html
<md-icon icon-set="fa">fa-home</md-icon>
```

which will become:

```html
<i class="md-icon fa fa-home"></i>
```

As an alternative, we could even leave the text in the `<i>` tag, and have it render as such:

```html
<i class="md-icon fa fa-home"><span style="display: none">fa-home</span></i>
```

Which gets some of the ligature benefits.

marcosmoura:
@Morgul The only problem with this approach is that the content of a component, in a slot, is not reactive, so we can't use like that. We will need to evaluate as a class. Also the icon-set should be properly prefixed with `md-`, `md-iconset` or `md-icon-set`.

Morgul:
@marcosmoura Ah, didn't realize it wasn't reactive. That's unfortunate.  (Also, I just missed the prefix, was unintended.)

In that case, I'm very happy with the proposal. :)

marcosmoura:
@Morgul Cool!
@sombriks Do you think that you can work with that?

sombriks:
@marcosmoura will be a pleasure!

marcosmoura:
Cool! Thank you a lot!

PR #293, opened 2017-01-05
fix a problem causing empty content in mdTable with pagination

meanevo:
this bug can be reproduced by:
1. create a table with pagination
2. fill some content in ( >mdSize so we can get multiple pages )
3. change page, for example, goto page 6 of total 7
4. switch to a larger RPP ('row per page') value, for example, from 10 to 25
5. it will result in a blank table, cause page 6 with 10 RPP displays row 51-60, but when changing rpp, it will stuck at page 6, which displays from row 5 * 25 + 1 = 126 > 7 * 10 = 70 (total data)

and also, by commenting out row per page selector's mounted value assignment, it now selects to md-size value by default, instead choosing the first option.


src/components/mdTable/mdTablePagination.vue
```diff
@@ -45,26 +45,46 @@
...
+      from() {
+        return (this.currentPage - 1) * this.currentSize + 1;
+      },
+      to() {
+        return this.currentPage * this.currentSize;
+      },
+      toCorrected() {
+        return this.isLastPage ? this.mdTotal : this.to;
+      },
+      total() {
```
marcosmoura: Why you are using this computed value? This is unused, right?

src/components/mdTable/mdTablePagination.vue
```diff
@@ -32,7 +32,7 @@
       },
       mdTotal: {
         type: [Number, String],
-        default: 'Many'
```
marcosmoura: The `Many` word is the default text for Google Projects, like on Gmail pagination on the top of the inbox.

src/components/mdTable/mdTablePagination.vue
```diff
@@ -45,26 +45,46 @@
...
     methods: {
       emitPaginationEvent() {
         if (this.canFireEvents) {
-          const sub = this.currentPage * this.currentSize;
-
-          this.subTotal = sub > this.mdTotal ? this.mdTotal : sub;
-          this.$emit('pagination', {
-            size: this.currentSize,
-            page: this.currentPage
+          this.$nextTick(() => {
```
marcosmoura: Why using a $nextTick here? To make sure that everything is updated?

meanevo: yes, it was to ensure currentPage get updated after changing pageSize.

src/components/mdTable/mdTablePagination.vue
meanevo: oh, didn't know that, my mistake.

src/components/mdTable/mdTablePagination.vue
meanevo: yes, missed out modifying it, cleaned up in following commit

Eder87rh:
Why this Issue was closed? I have the same problem

PR #296, opened 2017-01-05
Added 'change' event to mdBottomBar.

Morgul:
Now we fire off an event when the active item changes, just like `mdTabs`. Additionally, this moves activation logic to `mdBottomBar`, instead of each `mdBottomBarItem`. It made the change easier to implement, and seems a bit cleaner, though that's subjective.

The primary motivation for this is to better match `mdTabs`, and to make `mdBottomBar` a bit easier to work with.


marcosmoura:
Thank you! Great change!

PR #308, opened 2017-01-10
Fix row count

pablohpsilva:
## Bug
It happened when I removed one or more rows from the table and clicked on the "select all" checkbox. The number shown was the original data array size, not the new one (after deleting some itens). Example:

```
nutrition = [1,2,3,4,5];
numberOfRows = 5;

nutrition.splice(1,1);

nutrition = [1,3,4,5];
numberOfRows = 5; // this is the bug. it should be 4, not 5.
```

## Fix

mdTable.vue

```
  watch: {
    data() {
      this.numberOfRows = this.data.length;
    }
  },
```

This will guarantee the numberOfRows is always up to date.


docs/src/pages/ui-elements/Layout.vue
```diff
@@ -104,6 +104,12 @@
               </md-table-row>
 
               <md-table-row>
```
marcosmoura: Oops. Code from another PR!

pablohpsilva:
Sorry @marcosmoura . I forgot to stash before creating a new branch.

pablohpsilva:
One more thing? I'm pointing this PR towards the develop branch. I read the documentation and it said we shouldn't target the master. Is all alright?

marcosmoura:
Yes. The master branch aims to be the current release. Unless the bug is catastrophic we should create branches from develop and point it to the develop too.

PR #320, opened 2017-01-11
Fix MdTable.vue numberOfRows bug

pablohpsilva:
## Bug
It happened when I removed one or more rows from the table and clicked on the "select all" checkbox. The number shown was the original data array size, not the new one (after deleting some itens). Example:

```
nutrition = [1,2,3,4,5];
numberOfRows = 5;

nutrition.splice(1,1);

nutrition = [1,3,4,5];
numberOfRows = 5; // this is the bug. it should be 4, not 5.
```

## Fix

mdTable.vue

```
  watch: {
    data() {
      this.numberOfRows = this.data.length;
    }
  },
```

This will guarantee the numberOfRows is always up to date.

### DIFF
--- a/src/components/mdTable/mdTable.vue
+++ b/src/components/mdTable/mdTable.vue
@@ -55,8 +55,6 @@
       if (this.parentCard) {
         this.parentCard.tableInstance = this;
       }
-
-      window.$vm = this.data;
     }
   };
 </script>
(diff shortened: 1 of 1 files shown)


marcosmoura:
Cool. Sounds good! Merged!

PR #328, opened 2017-01-11
Fixes #327. Disable pointer events on snackbar container

### LINKED ISSUE #327: SnackBar container blocks clicks
When a snackbar is up, the topmost snackbar container block click events from happening underneath.

This happens even for the blank space to the right and left of the snackbar.

I've got a PR coming in a minute to fix this.

korylprince:
This just disables pointer events on the invisible snackbar container so clicks can pass through.


marcosmoura:
That's great. Thanks!

PR #330, opened 2017-01-12
Develop

sombriks:
This pull request should solve the issue #287 . 
The final usage differs very little from the proposal and i'll happily discuss it.
Please review this PR when possible. Thank you.

### DIFF
--- a/docs/src/pages/components/Icon.vue
+++ b/docs/src/pages/components/Icon.vue
@@ -50,8 +50,7 @@
               <md-table-row>
                 <md-table-cell>md-iconset</md-table-cell>
                 <md-table-cell><code>String</code></md-table-cell>
-                <md-table-cell>The icon set font you'll use. For example, in 
-                  <code>class="fa fa-pencil"</code>, the <b>fa</b> is the intended value for md-iconset.</md-table-cell>
+                <md-table-cell>The font icon set. Example: <code>md-iconset="fa fa-pencil"</code> for font awesome.</md-table-cell>
               </md-table-row>
             </md-table-body>
           </md-table>
--- a/docs/src/pages/components/Icon.vue
+++ b/docs/src/pages/components/Icon.vue
@@ -47,6 +47,12 @@
                 <md-table-cell><code>String</code></md-table-cell>
                 <md-table-cell>The src of the image file. Can be SVG icons too!</md-table-cell>
               </md-table-row>
+              <md-table-row>
+                <md-table-cell>md-iconset</md-table-cell>
+                <md-table-cell><code>String</code></md-table-cell>
+                <md-table-cell>The icon set font you'll use. For example, in 
+                  <code>class="fa fa-pencil"</code>, the <b>fa</b> is the intended value for md-iconset.</md-table-cell>
+              </md-table-row>
             </md-table-body>
           </md-table>
         </api-table>
@@ -130,6 +136,18 @@
                 &lt;/md-layout&gt;
               &lt;/md-layout&gt;
             </code-block>
+          </div>
+        </example-box>
+
+        <example-box card-title="Font Icons">
+          <div slot="demo">
+            <!-- had a few trouble on making webpack accept font awesome therefore i'm using this cdn for the moment -->
+            <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"/>
+            <md-icon md-iconset class="fa fa-2x fa-pencil"></md-icon>
+          </div>
+
+          <div slot="code">
+
           </div>
         </example-box>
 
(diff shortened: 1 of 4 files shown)


docs/src/pages/components/Icon.vue
```diff
@@ -5,7 +5,7 @@
         <p>A system icon, or UI icon, symbolizes a command, file, device, or directory. System icons are also used to represent common actions like trash, print, and save.</p>
         <p>You can see the full list of icons on the <a href="https://material.io/icons/" target="_blank" rel="noopener">Material Icons</a> website. Replace the spaces in the icon name by underscores (for example, <code>arrow back</code> becomes <code>arrow_back</code>).</p>
         <p>Alternatively you can use your PNG or SVG icons using the <code>md-src</code> property.</p>
-
+        <p>The use of font icons are suported too, although this approach lacks a few accessibility support.</p>
```
marcosmoura: "The use of font icons without ligatures are suported too". Because the Google Icons Font is "font icons" too. :)

docs/src/pages/components/Icon.vue
```diff
@@ -47,6 +47,12 @@
                 <md-table-cell><code>String</code></md-table-cell>
                 <md-table-cell>The src of the image file. Can be SVG icons too!</md-table-cell>
               </md-table-row>
+              <md-table-row>
+                <md-table-cell>md-iconset</md-table-cell>
+                <md-table-cell><code>String</code></md-table-cell>
+                <md-table-cell>The icon set font you'll use. For example, in
```
marcosmoura: I think that `md-iconset` should be a String prop. To be used like that:
``` vue
<md-icon md-iconset="fa fa-home"></md-icon> // In case of FA for example
```
``` vue
<md-icon>home</md-icon> // In case of Material Icons
```

src/components/mdIcon/mdIcon.vue
```diff
@@ -3,6 +3,8 @@
 
   <md-image class="md-icon" :class="[themeClass]" :md-src="imageSrc" v-else-if="imageSrc"></md-image>
 
+  <i class="md-icon md-icon-font" aria-hidden="true" :class="[themeClass]" v-else-if="mdIconset"></i>
```
marcosmoura: Then here you don't need a else-if because your iconset will evaluate as a class:
``` vue
<i class="md-icon" :class="[themeClass, mdIconSet]" :aria-hidden="!!mdIconSet" v-else>
  <slot></slot>
</i>
```

src/components/mdIcon/mdIcon.vue
```diff
@@ -17,7 +19,8 @@
 
   export default {
     props: {
-      mdSrc: String
+      mdSrc: String,
+      mdIconset: Boolean
```
marcosmoura: And here the prop will become:
``` js
mdIconset: {
  type: String,
  default: 'material-icons'
}
```

src/components/mdIcon/mdIcon.scss
```diff
@@ -10,6 +10,10 @@ $icon-size: 24px;
   font-size: $size;
```
marcosmoura: The font-size is already here...

src/components/mdIcon/mdIcon.scss
```diff
@@ -10,6 +10,10 @@ $icon-size: 24px;
...
 }
 
+.md-icon-font {
```
marcosmoura: ... So you don't need to set again :)

marcosmoura:
I made some comments. What do you think?

docs/src/pages/components/Icon.vue
sombriks: Agree! i'll change it to better express and specify the lack of ligatures.

docs/src/pages/components/Icon.vue
sombriks: This this the tricky part. For some reason it's not working, and that's why i tried a boolean one. However i'll make it string again, and see what happens. Maybe it was a problem between the couch and the laptop, ;-)

src/components/mdIcon/mdIcon.vue
sombriks: got it

docs/src/pages/components/Icon.vue
marcosmoura: Hahahaha. Don't worry mate!

src/components/mdIcon/mdIcon.vue
sombriks: seems to work very very well:
image

```
<md-icon md-iconset="fa fa-home"></md-icon>
<md-icon md-iconset="fa fa-bars" class="md-size-2x md-primary"></md-icon>
<md-icon md-iconset="fa fa-users" class="md-size-3x md-accent"></md-icon>
<md-icon md-iconset="fa fa-clock-o" class="md-size-4x md-warn"></md-icon>
```

src/components/mdIcon/mdIcon.scss
sombriks: This one was because i had trouble on font awesome. 

Just now, comparing with ionicons, i realize that is a "bug" on the font awesome side. They wire the font size at the font declaration.

Solve the proper default icon size for font awesome shouldn't be vue-material problem, therefore let it as it was originaly :+1: .

sombriks:
cool, i just pushed and the newer changeset already came there. :smiley:

marcosmoura:
Cool! Approved!

Issue #319, opened 2017-01-11
md-speed-dial improvement


korylprince:
The new speed dial is great!

However it is missing a particular feature.

From the spec:

> The floating action button can fling out related actions upon press. The button should remain on screen after the menu is invoked. Tapping in the same spot should either **activate the most commonly used action** or close the open menu.

Currently you can't have the speed dial to open on click, but have the morphed button become a clickable action, it only works to close the speed dial.

If you're willing, I'd like to add this feature in a new branch.

korylprince:
As far as implementation, I'm thinking to either pass a custom event when a click happens with the speed dial open, or using a prop to pass an event handler. I'm leaning towards the prop method. 

What are your thoughts?

marcosmoura:
Hi @korylprince 

I didn't get it. Have you tried the `@click.native` modifier on the md-fab-trigger?

korylprince:
@marcosmoura So that didn't work (the click handler takes over the md-fab-trigger action), but putting a `@click.native` on the underlying md-icon totally works. Thanks for the clue! I'll post my working implementation for future users:

```
<md-speed-dial md-open="click" md-mode="fling" class="md-fab-bottom-right">
	<md-button class="md-fab" md-fab-trigger>
		<md-icon md-icon-morph @click.native="$router.push({name: 'deviceCreate'})">computer</md-icon>
		<md-icon>add</md-icon>
	</md-button>

	<md-button class="md-fab md-primary md-mini" @click="$router.push({name: 'modelCreate'})">
		<md-icon>grid_on</md-icon>
	</md-button>

	<md-button class="md-fab md-primary md-mini" @click="$router.push({name: 'userCreate'})">
		<md-icon>account_circle</md-icon>
	</md-button>
</md-speed-dial>
```

marcosmoura:
Please note that `md-open="click"` `md-mode="fling"` are the default options, so you can omit them. :)

korylprince:
Fixed! Thanks.

korylprince:
I spoke too soon. This doesn't actually fix the problem.

It appears to possibly be a race condition, or the click event on the icon only works on a very small area.

I'm not sure what exactly is happening, but it very rarely works on the first try. You can click very rapidly and move the mouse over the Fab, and eventually it will do the native click handler, but it is by far a working solution.

Therefore I'll work up a PR for you to check out.

PR #331, opened 2017-01-12
Implement #319

korylprince:
I've played around with different ideas to implement what I wrote about in #319. The idea I chose is actually very simple, but I'm not sure what you'll think about it.

Firstly the idea is to be able to click the main fab in a speed dial when it's open and have it execute an action. Currently this can't work if md-open is click.

My idea is simply to have the speed dial component emit a `click` (or maybe you want to call it something more descriptive like `fab-trigger-click`, but that seems a bit long...) event if the speed dial is open and the fab-trigger is clicked.

You can listen for this event on the speed dial and execute whatever action you want (e.g. go to another route, open a dialog, etc)

What do you think? If you're okay with this I can update the docs with this event if you want.


pablohpsilva:
I think its a good idea. Wouldn't be also good to have where the user clicked?

korylprince:
@pablohpsilva passing the click event would require modifying the `toggleSpeedDial` signature, so I'll wait on @marcosmoura as to whether we should do that though.

To be clear though, this only fires when the fab-trigger md-button is clicked, so you basically already know where the user clicked...

marcosmoura:
I think that fire another event is better! Instead of @click. This might be confusing, because custom components need the .native modifier.

korylprince:
@marcosmoura , are you okay with renaming the event to `fab-trigger-click`?

korylprince:
@marcosmoura I pushed a second commit to use a custom event, `md-fab-click` (so it will be prefixed.)

If this doesn't work, can we discuss (perhaps on gitter) a solution to #319?

marcosmoura:
I don't know yet. I'm focusing on those bugfixes and then I get back here with an idea, ok?

korylprince:
No worries. I'm running my application from my fork right now. Fixing bugs is much better :)

Morgul:
@korylprince can you rebase your branch?

PR #347, opened 2017-01-13
Fix table arrow sort

jonataswalker:
Since we are going to rotate the arrow when sort is descending:

```css
.md-table .md-sortable.md-sorted-descending .md-sortable-icon {
  transform: translateY(-50%) rotate(180deg);
}
```

The right starting arrow is `arrow_upward`.


marcosmoura:
I'm sorry to take such a long time to answer that. I'm going to wait a bit more before merging this one. I will change the md-table a little, so I prefer to do all at once.

capttrousers:
@marcosmoura Am I correct in thinking the actual sorting functionality of the tables needs to be done outside of the md vue components? so clicking the column headers will trigger a custom sort function that sorts the items that make up the table rows?

PR #386, opened 2017-01-23
Feature/table improvements

gustavoBrid-IT:
Hi,

  At our company we decided to use Vue Material as frontend framework. Our actual basic needs was attended except by two components: a data table, more read-to-use and configurable, and a select component with server-side option, like select2 (next development).

  So we build a <md-data-table> component, that uses mdTable and others, but add some functionalities and facilities. Now we can fetch data from client or server, with local or remote pagination and more.

  I think it can be helpfull to others, so I'm submiting this pull request, what do you think? Is something to be on Vue Material core?

Regards!

### DIFF
--- a/docs/src/App.vue
+++ b/docs/src/App.vue
@@ -144,6 +144,10 @@
                 </md-list-item>
 
                 <md-list-item class="md-inset">
+                  <router-link exact to="/components/table/data-table">Table (DataTable)</router-link>
+                </md-list-item>
+
+                <md-list-item class="md-inset">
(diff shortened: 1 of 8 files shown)


marcosmoura:
Hi @gustavoBrid-IT 

Thank you for this PR, but I need to decline. This is not something that I would integrate with Vue Material core. I think you should provide this code by yourself in another repo, as a plugabble third-part component.

This will bring more complexity and add another library only to deliver this particular solution.

Thank you again.

PR #387, opened 2017-01-24
Update mdList.scss

warent:
Fix `md-list-item` with an `href` attribute causing the `.md-list-item-holder` to be off-center


marcosmoura:
I'm going to decline this PR. I'm creating (and almost finishing) a brand new mdList component, with the same behaviour and backward compatibility.

PR #429, opened 2017-02-02
Implement swipe on Sidenav

### LINKED ISSUE #275: Swipe on Sidenav
I got it to work, still need to make it more flexible and think about a solution for multiple sidenavs. But first, do you want it?

ezgif com-447b40bb5e

igor-ribeiro:
Fix #275

### DIFF
--- a/docs/src/pages/components/Sidenav.vue
+++ b/docs/src/pages/components/Sidenav.vue
@@ -18,19 +18,19 @@
 
             <md-table-body>
               <md-table-row>
-                <md-table-cell>swipeable</md-table-cell>
+                <md-table-cell>md-swipeable</md-table-cell>
                 <md-table-cell><code>Boolean</code></md-table-cell>
                 <md-table-cell>Enable the swipe functionality. Default <code>false</code></md-table-cell>
               </md-table-row>
 
               <md-table-row>
-                <md-table-cell>swipe-threshold</md-table-cell>
+                <md-table-cell>md-swipe-threshold</md-table-cell>
                 <md-table-cell><code>Number</code></md-table-cell>
                 <md-table-cell>Set the initial threshold for the swipe when it's closed. Default <code>15</code></md-table-cell>
               </md-table-row>
 
               <md-table-row>
-                <md-table-cell>swipe-distance</md-table-cell>
+                <md-table-cell>md-swipe-distance</md-table-cell>
                 <md-table-cell><code>Number</code></md-table-cell>
                 <md-table-cell>Set the swipe distance needed to open/close the sidenav. Default <code>100</code></md-table-cell>
               </md-table-row>
@@ -132,7 +132,7 @@
                 <p>Open console to see the events</p>
               </div>
 
-              <md-sidenav class="md-left" ref="leftSidenav" @open="open('Left')" @close="close('Left')" swipeable>
+              <md-sidenav class="md-left" ref="leftSidenav" @open="open('Left')" @close="close('Left')" md-swipeable>
                 <md-toolbar class="md-large">
                   <div class="md-toolbar-container">
                     <h3 class="md-title">Sidenav content</h3>
--- a/docs/src/pages/components/Sidenav.vue
+++ b/docs/src/pages/components/Sidenav.vue
@@ -7,6 +7,36 @@
 
       <div slot="api">
         <api-table name="md-sidenav">
+          <md-table slot="properties">
+            <md-table-header>
+              <md-table-row>
+                <md-table-head>Name</md-table-head>
+                <md-table-head>Type</md-table-head>
+                <md-table-head>Description</md-table-head>
+              </md-table-row>
+            </md-table-header>
+
+            <md-table-body>
+              <md-table-row>
+                <md-table-cell>swipeable</md-table-cell>
+                <md-table-cell><code>Boolean</code></md-table-cell>
+                <md-table-cell>Enable the swipe functionality. Default <code>false</code></md-table-cell>
+              </md-table-row>
+
+              <md-table-row>
+                <md-table-cell>swipe-threshold</md-table-cell>
+                <md-table-cell><code>Number</code></md-table-cell>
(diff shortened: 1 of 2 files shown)


igor-ribeiro:
I will change to Mouse Events [http://caniuse.com/#feat=touch]

igor-ribeiro:
Actually, it looks like there are no problems on mobile support, so I guess we can leave it like that.

docs/src/pages/components/Sidenav.vue
```diff
@@ -7,6 +7,36 @@
...
+              <md-table-row>
+                <md-table-head>Name</md-table-head>
+                <md-table-head>Type</md-table-head>
+                <md-table-head>Description</md-table-head>
+              </md-table-row>
+            </md-table-header>
+
+            <md-table-body>
+              <md-table-row>
+                <md-table-cell>swipeable</md-table-cell>
```
marcosmoura: The attributes should be prefixed with "md-" :)

docs/src/pages/components/Sidenav.vue
```diff
@@ -7,6 +7,36 @@
...
+                <md-table-cell><code>Boolean</code></md-table-cell>
+                <md-table-cell>Enable the swipe functionality. Default <code>false</code></md-table-cell>
+              </md-table-row>
+
+              <md-table-row>
+                <md-table-cell>swipe-threshold</md-table-cell>
```
marcosmoura: Same here.

docs/src/pages/components/Sidenav.vue
```diff
@@ -7,6 +7,36 @@
...
+                <md-table-cell><code>Number</code></md-table-cell>
+                <md-table-cell>Set the initial threshold for the swipe when it's closed. Default <code>15</code></md-table-cell>
+              </md-table-row>
+
+              <md-table-row>
+                <md-table-cell>swipe-distance</md-table-cell>
```
marcosmoura: And here.

docs/src/pages/components/Sidenav.vue
```diff
@@ -102,7 +132,7 @@
                 <p>Open console to see the events</p>
               </div>
 
-              <md-sidenav class="md-left" ref="leftSidenav" @open="open('Left')" @close="close('Left')">
+              <md-sidenav class="md-left" ref="leftSidenav" @open="open('Left')" @close="close('Left')" swipeable>
```
marcosmoura: Same here too.

src/components/mdSidenav/mdSidenav.vue
```diff
@@ -45,7 +56,71 @@
         } else {
           this.open();
         }
+      },
+      handleTouchStart(event) {
+        const positionX = event.touches[0].clientX - this.mountedRect.left;
+        const positionY = event.touches[0].clientY - this.mountedRect.top;
+
+        if (positionX < 0 || positionX > this.mountedRect.left + this.mountedRect.width) {
```
marcosmoura: Maybe will be good to create three methods with the conditions inside those if statements, and then merge then to only one if, like that:
 
``` js
if (this.methodOne() || this.methodTwo() || this.methodThree()) {
  return;
}
```

src/components/mdSidenav/mdSidenav.vue
```diff
@@ -45,7 +56,71 @@
...
+
+      this.mountedRect = this.$refs.backdrop.$el.getBoundingClientRect();
+      this.initialTouchPosition = null;
+      this.canMove = false;
+
+      document.addEventListener('touchstart', this.handleTouchStart);
+      document.addEventListener('touchend', this.handleTouchEnd);
+      document.addEventListener('touchmove', this.handleTouchMove);
+    },
+    destroyed() {
```
marcosmoura: Instead of destroyed, move this to beforeDestroy

marcosmoura:
GREAT job ever here. Just created some comments to make it more compliant with the code style and standards.

src/components/mdSidenav/mdSidenav.vue
igor-ribeiro: Is this OK for you?
```js
if (!this.isHorizontallyInside(positionX) || !this.isVerticallyInside(positionY) || !this.isFromStartWhenClosed(positionX)) {
  return;
}
```

igor-ribeiro:
The only problem is that we cannot swipe from right to left to open. I can work on it if you want.

marcosmoura:
@igor-ribeiro I think, for now, we don't need to develop this. I will create some components to the the "app-shell". And then will be easier to create ths behaviour. I will take a deep look at this code and I get back to you. Thanks a lot for this PR.

fergardi:
Any idea why this suddenly stopped working for me? Was running fine in my Cordova app but no more swipe effect... :(

igor-ribeiro:
@fergardi Can you detail a little more? I work with someone that is using it and the swipe is working. Try to adjust the props `md-swipe-distance` and `md-swipe-threshold`.

fergardi:
I added it but no luck yet. The milestone for this was changing from default cordova webview to cordova crosswalk webview plugin (latest chromium canary). Before that it worked like a charm. Now thanks to that plugin it seems to load and render really really fast in comparison, specially the images and scrolling behaviour, but in exchange I lost the swipe events (and probably other events) in the process. Not working anymore. At all. Maybe has to do with something related to the new @click.native events? Dunno. For now, I'll keep the xwalk plugin as it speeds up my app like a LOT, but would love to have back the swipe events. Could you add this plugin to a cordova project and test your voodoo magic against it?

And btw, could you point me in the right direction to add your md-swipeable functionality by default to my new component in  #549? Would be REALLY cool. Like... mandatory.

Thanks for your effort.

fergardi:
Never mind, It somehow fixed itself. Maybe was an outdated plugin messing touch events. Btw I added your functionality to my new slider component!

Linzdigr:
Would love to have a simple way to get the right sidenav working with swipe! (instead of the default left to right). Nice job anyway.

Issue #267, opened 2016-12-28
Are tabs head swipable?


zapping:
Hi,
 Can you please take a look at this codepen https://codepen.io/zapping/pen/bBPyOz . 

 I tried added the tabs component. But when the no. of tabs increases it does not seem to have a scroll option on desktop browser nor can the tabs be swipped on mobile.

 Have I missed/overlooked something? Can you please clarify on how the tab heads can be navigated in such a case i,e when you have lots of tabs. It will be preferable to have the swipe option.

Thanks and you have a great set of components.

marcosmoura:
This is not possible yet! I will develop this in the next versions.
For now, a possible workaround on mobile devices, is to add a custom css:
``` scss
.md-tabs-navigation {
  overflow: hidden;
}
```

Sorry about that. Thank you.

zapping:
np. Thank you for the work around.

igor-ribeiro:
I got it to work, only problem is the indicator.

igor-ribeiro:
Got it!

![tabs-header-scroll](https://cloud.githubusercontent.com/assets/5148753/22560552/311c721c-e95c-11e6-9a69-610666d3c9fd.gif)

zapping:
Oh, thats nice!! Does it mean that this will be available on the release sooner, thx.

marcosmoura:
@zapping Yep!

zapping:
Hi Marcos,
Can you please take a look at this pen, http://codepen.io/zapping/pen/bBPyOz , and let me know what I missed? The add button on top needs to be clicked twice for the first tab to appear and also the tab needs to be clicked for the content to appear. Same is the case with the delete button too.

Is there a way to customize the tab headers so that a small delete button can be added to each tab header to close it?

When I try to change the Add button to <md-button @click="addTab()">Add</md-button> the method of @click is not being called for some reason.

Thanks.

zapping:
Hi Marcos,
 Please ignore the last one, realized that @click.native="addTab()" is the way to go.

Thanks.

zapping:
Hi Marcos,
 Can you please check the first 2 comments on the post https://github.com/marcosmoura/vue-material/issues/267#issuecomment-280055906

Thanks.

igor-ribeiro:
[...]

PR #430, opened 2017-02-02
Implement swipe on Tabs header

igor-ribeiro:
#267


marcosmoura:
I think this option should be default and applied anyway. What do you think?

marcosmoura:
@igor-ribeiro Can you change that?

igor-ribeiro:
@marcosmoura I will!

igor-ribeiro:
I did it. The only problem is when you resize it on desktop the scrollbar becomes visible. This happens because my hack to hide the scrollbar uses negative margin, but on mobile there is no scrollbar like on desktop, so there is no need for the negative margin.

src/components/mdTabs/mdTabs.vue
```diff
@@ -40,6 +42,7 @@
       mdFixed: Boolean,
       mdCentered: Boolean,
       mdRight: Boolean,
+      mdScrollableNavigation: Boolean,
```
marcosmoura: Can you make this property true by default?

dohomi:
It would be great if the tab header would have a left/right navigation arrows visible here: https://material.angularjs.org/latest/demo/tabs. Thanks for this improvement!

src/components/mdTabs/mdTabs.vue
igor-ribeiro: I can actually removed since it is the default behavior now.

igor-ribeiro:
It would be very nice. What do you think, @marcosmoura ?

Morgul:
@igor-ribeiro with regards to the arrows, I think they should clip the tabs under them; the transparency is a little hard to read. But _personally_, I really like that idea. I've wanted that functionality a few times in a tab component.

igor-ribeiro:
@Morgul I agree!

igor-ribeiro:
tabs-navigation

marcosmoura:
I think this is awesome, but can you create a different PR for that?

marcosmoura:
@igor-ribeiro Can you sync your branch with the latest develop?

marcosmoura:
@igor-ribeiro ping

igor-ribeiro:
@marcosmoura Done!

marcosmoura:
Sounds good!
I'm just holding this until v0.8.0 release.
Thank you so much for this PR.

sinisarudan:
Hi, is this still not solved?
Namely, even after adding the prop:
<md-tabs md-swipeable>, md-tabs looks like this.
tabs being cut-out, squeezed and not swipeable.
image

PR #435, opened 2017-02-03
Allow preventDefault in md-switch

ivictbor:
This allows call `$event.preventDefault()` from `<md-switch @change=''`. It is usefull if we want to prevent changing of switch state to confirm change in dialog. Also issue related to this https://github.com/marcosmoura/vue-material/issues/433
Possible use:
```html
<md-switch  @change="$event.preventDefault();start_confirm_dialog()">
```
Or same in `Vue` style:
```html
<md-switch  @change.prevent="start_confirm_dialog()">
```
Also it allows `$event.stopPropagation()` or `@change.stop` which is usefull if md-swith is placed on expanded list and if developer wants to prevent expanding of list when user his on switch.
Solution explained in comments. Thanks

### DIFF
--- a/src/components/mdSwitch/mdSwitch.vue
+++ b/src/components/mdSwitch/mdSwitch.vue
@@ -66,9 +66,22 @@
         this.leftPos = this.checked ? checkedPosition + '%' : initialPosition;
       },
       changeState(checked, $event) {
-        this.checked = checked;
-        this.$emit('change', this.checked, $event);
-        this.$emit('input', this.checked, $event);
+        if ($event !== undefined) {
+          // $event is not undefined when changeState called from toggle (user click)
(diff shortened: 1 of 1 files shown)


src/components/mdSwitch/mdSwitch.vue
```diff
@@ -66,9 +66,22 @@
         this.leftPos = this.checked ? checkedPosition + '%' : initialPosition;
       },
       changeState(checked, $event) {
-        this.checked = checked;
-        this.$emit('change', this.checked, $event);
-        this.$emit('input', this.checked, $event);
+        if ($event !== undefined) {
```
marcosmoura: If you want to test undefined, please test the type:
typeof $event !== 'undefined'

ivictbor: Thanks, done

PR #441, opened 2017-02-04
Option that allow multiple items be expanded in same time in md-list

ivictbor:
Option `allow-multiple-expanded` that allow multiple expanded items in `<md-list>`. Default value is `false` so it will not brake any backward compatibility.
* Behavior when `allow-multiple-expanded` or not specified (works as it was before): if some item already expanded, expanding new will fold already expanded 
* When `allow-multiple-expanded` is used (`<md-list allow-multiple-expanded>` ): if some item already expanded, expanding new will not fold already expanded, but click on expanded item still will fold it.

### DIFF
--- a/docs/src/pages/components/List.vue
+++ b/docs/src/pages/components/List.vue
@@ -38,6 +38,24 @@
               </md-table-row>
             </md-table-body>
           </md-table>
+
+          <md-table slot="properties">
+            <md-table-header>
+              <md-table-row>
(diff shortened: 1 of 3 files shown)


marcosmoura:
Hi @ivictbor,

I'm going to use the same logic that you create here in the new md-list component. So I will close this PR for now, but I will give credits for you in this one.

Thank you SO MUCH for doing this!

PR #442, opened 2017-02-04
Allows place expand indicator at the beginning of item

ivictbor:
Allows place expand indicator at the beginning of item (must have feature to design some interfaces)


marcosmoura:
Hi @ivictbor,

I'm going to use the same logic that you create here in the new md-list component. So I will close this PR for now, but I will give credits for you in this one.

Thank you SO MUCH for doing this!

Issue #344, opened 2017-01-13
MdInkRipple error in disabled watcher


bolerodan:
### Steps to reproduce
 Put a :disabled attribute on md-button and toggle it on and off. When putting disabled = true, everything is fine, md-ink-ripple is removed as per the code, however, putting :disabled to false will result in an error
```
TypeError: Cannot read property 'parentNode' of null
```
### Which browser?
Chrome 55

### What is expected?
for md-ink-ripple to append itself back to its initial parent (in this case, md-button)
### What is actually happening?
Fails to attach it self back to its parent

This may be related to  #329  as well.
Looking through the code, the error is in the watcher of the disabled prop. When disabled is true, md-ink-ripple is destroyed, removed from parent. When switched to true, `this.init()` is executed, however at that point in time,` this.$el` is no longer attached to its parent, thus when calling `getClosestPositionedParent(`) in the `init()` function, there is no parent on `this.$el`

Either in the watcher for disabled, `this.$el` should be reinserted into `this.parentElement` before calling init().

I wish I had time to setup a PR but I was able to investigate this much. Thanks again for your awesome work.

jonataswalker:
Related #343.

korylprince:
This was already reported in #313 and fixed in PR #317 (using pretty much your code.) It should make it into version 0.6.4.

bolerodan:
Maybe im reading the code wrong in PR #317 but does that seem to only do a null check and not re insert the md-ink element back into its parent? If I disable the the button, then re-enable it, the watcher for disabled will call init() because disabled will be false, then this.getClosestPositionedParent() with the new PR will return false thus not re-enabling md-ink ?

Perhaps I am reading that PR wrong.

marcosmoura:
I will take a deep look into it. Thanks!

korylprince:
@marcosmoura 

Currently in `init`, `this.parentElement` is always reset to `this.getClosestPositionedParent(this.$el.parentNode)`. Perhaps `this.parentElement` should only be set if it is not `null`.

I would do a PR, but obviously my last solution didn't fully solve the problem. I want to make sure I fully understand what's going on here.

copitz:
[...]

PR #449, opened 2017-02-07
mdInkRipple disabled fix

bolerodan:
So while the previous PR for issue #344 silences the undefined error that was being thrown in getClosestPositionParent, it does not fix the issue at hand.

If disabled is set to true, the watcher calls this.destroy() removing the rippleElement from the parent. When disabled is set to false, the watcher calls this.init(). However at that point, rippleElement is no longer in the DOM, thus causing getClosestPositionedParent to throw the error originally.

The new fix currently just silences it, and currently in develop, does not re-append the rippleElement so when you toggle between disabled true/false, the rippleElement is never re-activated.

This checks in the watcher, and re-appends the rippleElement, however there are edge-cases.

If the button starts disabled, references to the parentElement are not caught because they are called in init(), and mounted checks for disabled, thus a null would be thrown in the watcher. I had to remove those out of init and set them in mounted so references are always stored for edge cases.

However another problem arises in the template. If disabled is set to true to begin with, mounted does not have the original DOM element, thus causing issues with stored references, so I removed the v-if from the template. I'm not 100% sure why that was there originally.

Take a look at this. I hope this brings up the original issue at hand and you can choose how to fix it.

Thanks a lot for you work.


marcosmoura:
This was already fixed on v0.7.0. If something appear back I will take a deep look at this solution. Thanks a lot for this PR! :D

bolerodan:
@marcosmoura  Are you sure? The code that this PR is against, hasnt changed between my PR and the release of 0.7.0

I will put up a fiddle soon. All you need to do, is toggle the disable for an md-button (or md-ink actually). So disable it, then Re-enable it. you will notice that md-ink-ripple no longer works.

The original PR technically did not solve what was actually happening, it just covered up the error ;)

TL;DR if you disable a button, then re-enable it, md-ink-ripple no longer works (but doesnt crash)

bolerodan:
@marcosmoura   Here is a pen example

http://codepen.io/anon/pen/KaGpLg?editors=1111

click button. Ripple works. Click the checkbox to disable. Uncheck the checkbox to re-enable. Click button, ripple no longer works.

terpret:
Same here.

PR #466, opened 2017-02-08
[fix] @click.native on dialog presets

jtouzy:
Fix dialogs since 0.7.0


jtouzy:
Already done in fe1d3ce977fb49d77db8b6301895c9dc9ef02088.

PR #473, opened 2017-02-09
Add md-clearable to md-input-container

d3radicated:
Using `type="search"` presents you with a generic clear button. This adds a button (like "md-has-password") which when clicked, clears the input.

### DIFF
--- a/src/components/mdInputContainer/mdInputContainer.vue
+++ b/src/components/mdInputContainer/mdInputContainer.vue
@@ -86,7 +86,8 @@
       clearInput() {
         if (this.isInput()) {
           this.input.value = '';
-          this.input.dispatchEvent(new Event('input'));
+          this.setValue(this.input.value);
+          this.$emit('input', this.input.value);
         }
       },
       setValue(value) {
--- a/src/components/mdInputContainer/mdInputContainer.vue
+++ b/src/components/mdInputContainer/mdInputContainer.vue
@@ -8,7 +8,7 @@
       <md-icon>{{ showPassword ? 'visibility_off' : 'visibility' }}</md-icon>
     </md-button>
 
-    <md-button class="md-icon-button md-clear-input" @click="clearInput" v-if="mdClearable && hasValue">
+    <md-button class="md-icon-button md-clear-input" @click.native="clearInput" v-if="mdClearable && hasValue">
       <md-icon>clear</md-icon>
     </md-button>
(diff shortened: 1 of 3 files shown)


src/components/mdInputContainer/mdInputContainer.vue
```diff
@@ -77,6 +83,12 @@
           this.input.focus();
         }
       },
+      clearInput() {
+        if (this.isInput()) {
+          this.input.value = '';
+          this.input.dispatchEvent(new Event('input'));
```
marcosmoura: This use the vue syntax for that: this.$emit('input', this.input.value)

src/components/mdInputContainer/mdInputContainer.vue
```diff
@@ -77,6 +83,12 @@
           this.input.focus();
         }
       },
+      clearInput() {
+        if (this.isInput()) {
+          this.input.value = '';
```
marcosmoura: You should call `this.setValue(this.input.value)` to fix label and placeholder.

marcosmoura:
Hi @d3radicated 

Thank you so much for this PR.
Can you post an screenshot of this change?
Also, it's missing an example on Input.vue. Can you add please?

d3radicated:
Here's a preview of the component.

[Image](https://i.imgur.com/tiUphsT.png)

Regarding this change: `this.$emit('input', this.input.value);` It doesn't fire the 'input' event. That's why I used `dispatchEvent` to fire the event.

src/components/mdInputContainer/mdInputContainer.vue
```diff
@@ -7,6 +7,10 @@
     <md-button class="md-icon-button md-toggle-password" @click.native="togglePasswordType" v-if="mdHasPassword">
       <md-icon>{{ showPassword ? 'visibility_off' : 'visibility' }}</md-icon>
     </md-button>
+
+    <md-button class="md-icon-button md-clear-input" @click.native="clearInput" v-if="mdClearable && hasValue">
```
marcosmoura: I think the button should be highlighted with the primary color when the input receives focus. An easy way to do that is to toggle the 'md-primary' class when `this.isFocused` is true.

d3radicated: It already does. It has the same style of `md-password`.

PR #474, opened 2017-02-09
Scroll the selected md-menu into view when the md-select is opened.

qorsmond:
This is my stab at a large list of items in the md-select when the user selects a option towards the end of the list and reopens the list it should automatically scroll to the selected item.

For example when a user selects the Verdana font and reopens the md-select it should show that selection.

image


marcosmoura:
Hi @qorsmond 
Thank you so much for this PR. I decided to develop this in another way, but I will totally give you the credits for this one. :)

PR #485, opened 2017-02-10
Added Grid component

fergardi:
Hello.

Wanted to collaborate again. What about this?

image

Cheers!


marcosmoura:
I liked the idea, but there's some points missing. 
* The grid should have ripple by default
* Instead of md-grid-media, a md-grid-cell name would be more clear
* The cell should use <md-image> internally. So maybe instead of let the user to use real  tags, we can accept a md-src to point to the image
* Every internal class should be prefixed with `md-`
* I dont think the single-line/double-line classes are needed. We can support both that using flexbox.
* We need to support md-layout by default and internally. If the resize occurs, we need to rearrange the items equally to fill the spaces. And we need to support nice animations, like on "Masonry" plugin, but without using it inside the library. This part is the most difficult one.
* Don't use custom media queries. There is some breakpoints set in the variables.scss file. You can use the example that is inside md-layout component.
* There is some "Stepper" contents inside this branch

Do you think that you can implement them?

marcosmoura:
I liked the idea, but there's some points missing. 
* The grid should have ripple by default
* Instead of md-grid-media, a md-grid-cell name would be more clear
* The cell should use <md-image> internally. So maybe instead of let the user to use real  tags, we can accept a md-src to point to the image
* Every internal class should be prefixed with `md-`
* I dont think the single-line/double-line classes are needed. We can support both that using flexbox.
* We need to support md-layout by default and internally. If the resize occurs, we need to rearrange the items equally to fill the spaces. And we need to support nice animations, like on "Masonry" plugin, but without using it inside the library. This part is the most difficult one.
* Don't use custom media queries. There is some breakpoints set in the variables.scss file. You can use the example that is inside md-layout component.
* There is some "Stepper" contents inside this branch

Do you think that you can implement them?

fergardi:
I'll give it a complete and cleaner rework.

pablohpsilva:
@fergardi this is gold!
Can you make the changes that @marcosmoura said? I'm crazy to merge this PR.

fergardi:
Sorry but I switched to http://www.muse-ui.org/#/gridList.

PR #495, opened 2017-02-13
Fix problem with german keyboards.

matthinc:
The key 186 is the 'ö' key on german keyboards. I'm sure it means many other things on many other keyboard. So this line would break the component for many people and should be removed.


marcosmoura:
I will accept this temporarily. I will change the component to make it flexible enough to pass custom keycodes to the component. Thank you for this PR.

Issue #497, opened 2017-02-13
Themed MdChips


fergardi:
I know MD specs says that all chips should be lightgrayed, but we have so nice theming customization right now it would be a shame not to use it on all elements, including MdChips, dont you think? A small theme.scss could be added to inherit the .md-primary .md-accent and md-warn to the chips.

Just my 2cents.

fergardi:
Resolved with #498.

PR #498, opened 2017-02-13
Themed Chips

fergardi:
Added themed Chips.

#497 

image

What do you think?

### DIFF
--- a/docs/src/pages/components/Chips.vue
+++ b/docs/src/pages/components/Chips.vue
@@ -223,6 +223,24 @@
             </code-block>
           </div>
         </example-box>
+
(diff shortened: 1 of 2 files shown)


fergardi:
Done.

docs/src/pages/components/Chips.vue
```diff
@@ -31,6 +31,30 @@
             </md-table-body>
           </md-table>
 
+          <md-table slot="classes">
```
fergardi: Done.

d3radicated:
I saw this and I love this feature. Is there a way to integrate this with `md-chips`? Like all of the inputted chips there will have a theme.

PR #512, opened 2017-02-15
md-editable property for md-chip

albanm:
This PR adds a 'md-editable' property  and 'edit' event on md-chip.

They are then used by md-chips to support editing the chips. When a chip is clicked it is removed from the list, and it's content is added to the input where the user can modify it then add it again.

I think it is quite straightforward and integrates nicely with the existing features. I hope you will agree.

On a side note: I wrote my name in the new chip example. I did it because of my lack of imagination, nothing else, feel free to write whatever you want :)


marcosmoura:
Man. Nice solution. I liked. And merged!
<sub><sup>Your name will stay there! :D

PR #517, opened 2017-02-15
Fix for issue with md-table-pagination component not enabling the paging buttons when total items is updated, after component is rendered

praneetloke:
- Added watchers on mdTotal and mdPage. 
- Changed the binding for the disabled attribute of the next page arrow to use a computed property.

This PR is for the bug described in #482.

PR #521, opened 2017-02-16
Follow-up for #517

praneetloke:
@marcosmoura this is a follow-up of PR #517 to change the name of the computed property for disabling the next-page button; this is to make the code a bit more clear and easy to read. 

I have also changed the `watch` handlers to use the value passed to them instead of referencing the values from the component instance for `mdSize` and `mdPage`.

P.S: I merged your master branch to my fork before raising this PR.


marcosmoura:
Sounds great. I prefer those names too. Thank you!

Issue #492, opened 2017-02-12
 [mdInput] calculated height on start


delueg:
Hello,

when the md-textarea already has a set value from v-modal the height of the textarea will not calculate correctly.

https://jsfiddle.net/rgdoyjcs/

is it possible to trigger the height calculation?

marcosmoura:
It's definitely a bug. I will fix that in the next version.

leMaur:
I got an error about textarea when I minify the code! 
`Uncaught TypeError: Cannot read property 'height' of undefined at HTMLTextAreaElement.s`

My setup is:
- Laravel 5.3 
- elixir 6.0
- elixir-webpack-official 1.0.2
- Vue.js 2.0.1
- VueMaterial 0.7.1

Someone had the same problem?
I'm not sure is VueMaterial md-textarea component instead the autosize plugin.
Any idea?

P.S.
Thank you @marcosmoura for this great project!

jdrydn:
Looking forward to this being fixed 😄  Only (noticeable) bug in my new CMS rebuild is this 🎉

Samuell1:
Closing this issue as our focus is on the new 1.0.0 version.

PR #525, opened 2017-02-17
Automatically adjust the textarea height when rendering the model

jdrydn:
As per #492, here's a small fix to automatically adjust the height of the textarea. I'm not too pleased with the implementation, but I can't find a nicer way to do it. Not sure if this is actually a bug in `autosize` or not?


jdrydn:
(Also, my apologies, I branched off the `master` branch and set the PR to merge into develop (because that's what your PR guidelines requested) 😉

PR #526, opened 2017-02-17
Make the label interact with the checkbox even if no name or id property is set

jdrydn:
Whilst using this (excellent) library, I found clicking on a `<label></label>` of a `<md-checkbox></md-checkbox>` component never actually checked the checkbox if there was no `name` or `id` property set. To fix this I added the Vue UID of the instance to `md-checkbox` every time, and I figured it made sense to make the checkbox default back to `this._uid` if no name or id is set 😉

### DIFF
--- a/src/components/mdCheckbox/mdCheckbox.vue
+++ b/src/components/mdCheckbox/mdCheckbox.vue
@@ -1,11 +1,11 @@
 <template>
   <div class="md-checkbox" :class="[themeClass, classes]">
     <div class="md-checkbox-container" @click.stop="toggleCheck" tabindex="0">
-      <input type="checkbox" :name="name" :id="id" :disabled="disabled" :value="value" :checked="checked" tabindex="-1">
+      <input type="checkbox" :name="name" :id="id || this._uid" :disabled="disabled" :value="value" :checked="checked" tabindex="-1">
       <md-ink-ripple :md-disabled="disabled" />
     </div>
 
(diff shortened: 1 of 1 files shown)


jdrydn:
(Also, my apologies, I branched off the `master` branch and set the PR to merge into develop (because that's what your PR guidelines requested) 😉

Samuell1:
`id, for` its not needed if its input is inside label

jdrydn:
@Samuell1 Agreed. If you look at the original code you'll see the input is not inside the label.

Samuell1:
@jdrydn yes, but it will be cool

src/components/mdCheckbox/mdCheckbox.vue
```diff
@@ -1,11 +1,11 @@
 <template>
   <div class="md-checkbox" :class="[themeClass, classes]">
     <div class="md-checkbox-container" @click.stop="toggleCheck" tabindex="0">
-      <input type="checkbox" :name="name" :id="id" :disabled="disabled" :value="value" :checked="checked" tabindex="-1">
+      <input type="checkbox" :name="name" :id="id || this._uid" :disabled="disabled" :value="value" :checked="checked" tabindex="-1">
```
marcosmoura: @jdrydn The _uid is an integer that can colide with others IDs in the same page. There is a uniqueId generator in this project. It was used in the tabs component. Could you change this?

PR #531, opened 2017-02-17
Core/components/md theme

cawa-93:
Add `textColor` property to theme options


marcosmoura:
@cawa-93 There is some code of md-table. Can you remove it?

PR #535, opened 2017-02-20
Fix - when to use default colorVariant

vizo:
Accent and background hue was always set to default value.

Solves issue https://github.com/marcosmoura/vue-material/issues/533.


DavisMiculis:
Yes, please accept this! 👍

marcosmoura:
There is some code related to md-table. Can you remove it?

PR #541, opened 2017-02-20
Added v-on:click (@click) support to md-button

u12206050:
Since by default a button usually has a click event I feel it is by standard useful to have the native click event sent up via a custom event conforming to what most would expect to work.
Used `v-on:click.stop` to stop the event from causing issues and conflicts between multiple other events
`<md-button @click="alert('Hello World')"></md-button>`
instead of 
`<md-button @click.native="alert('Hello World')"></md-button>`

### DIFF
--- a/src/components/mdButton/mdButton.vue
+++ b/src/components/mdButton/mdButton.vue
@@ -1,5 +1,5 @@
 <template>
-  <button class="md-button" :class="[themeClass]" :type="type" :disabled="disabled" v-if="!href">
+  <button class="md-button" :class="[themeClass]" :type="type" :disabled="disabled" v-if="!href" v-on:click.stop="onClick($event)">
     <md-ink-ripple :md-disabled="disabled"></md-ink-ripple>
     <slot></slot>
   </button>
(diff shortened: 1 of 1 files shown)


jtouzy:
From 0.7.0 changelog :

> Breaking changes:
> To avoid issues and conflicts between multiple events on buttons and other "clickable" components the internal @click event don't work anymore. Use @click.native instead.

Morgul:
@marcosmoura I feel like this is a design that needs more discussion. Vue Material's change to force `@click.native` lead me to question my design choice of reemitting in some of the libraries I've been writing, and left me wondering if I was not following best practices, so I asked on the Vue forums:

https://forum.vuejs.org/t/should-libraries-require-native/7885

Evan responded with this:

> I would personally re-emit it to present a more natural interface to the user. The only possible drawback of re-emitting is you need to be explicit about the events you re-emit in the docs so that the users won't be surprised when they try some other events without `.native` and find them not working.

(There was also a mention of making those components [functional][], where that paradigm makes sense.)

Personally, I find the change to `@click.native` to be a very frequent papercut, as none of the other libraries I use frequently require it, and it's not how I've trained myself to write Vue code.

[functional]: https://vuejs.org/v2/guide/render-function.html#Functional-Components

jonataswalker:
I share @Morgul thoughts.

marcosmoura:
This makes sense. I will decline this PR and make this change globally!

PR #545, opened 2017-02-21
Fix registering/unregistering tabs

igor-ribeiro:
Vue cannot detect properly setting and removing object properties (https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats).

#267


marcosmoura:
Man. Awesome! Thank you!

PR #578, opened 2017-02-28
md-error possibility and remove required chip-template within md-chips

romangit:
Hi, I would like remove required chip-template and make md-error possibility.


marcosmoura:
Awesome! Merged!

PR #583, opened 2017-03-02
checkbox: allow keyboard toggling

fiws:
pretty simple implementation of keyboard toggling

### DIFF
--- a/src/components/mdCheckbox/mdCheckbox.scss
+++ b/src/components/mdCheckbox/mdCheckbox.scss
@@ -19,6 +19,7 @@ $checkbox-touch-size: 48px;
 
     &:focus {
       outline: none;
+      box-shadow: 0 0 0 2px rgba(#000, .14);
     }
 
     &:before {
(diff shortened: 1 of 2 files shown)


src/components/mdCheckbox/mdCheckbox.scss
```diff
@@ -19,6 +19,7 @@ $checkbox-touch-size: 48px;
 
     &:focus {
       outline: none;
+      box-shadow: 0 0 0 2px rgba(#000, .14);
```
Morgul: Why is this included in this change? It doesn't seem related to keyboard toggling...

fiws: otherwise there will be no visual indication that the checkbox is focused. keyboard toggling is kinda hard without that ;)

Morgul: Ah! Ok, that makes sense. :)

marcosmoura: I think that instead of box-shadow we should highlight the md-ripple element inside. Do you think that you can change that?

fiws: so, i previously did that (at least made it look like that). 
Thing is: toggling the checkbox will not remove the focus even when clicking, so the active ripple will look rather strange.

PR #596, opened 2017-03-09
Allow customize tab item using slot

vanhoavn:
Allow customize tab's header item using slot in `md-tabs` component.


marcosmoura:
Man. Great PR and nice idea. Sounds good! Merged!

Issue #608, opened 2017-03-16
Enlarge mdRadio touchable area, like mdCheckbox


zcyzcy88:
The blank around of mdCheckbox is touchable:
> <img width="102" alt="checkbox" src="https://cloud.githubusercontent.com/assets/15178410/23999747/34b5cf78-0a94-11e7-8bc5-8a65f81e2e78.png">

But mdRadio is not:
> <img width="98" alt="radio" src="https://cloud.githubusercontent.com/assets/15178410/23999882/9c6cae5c-0a94-11e7-9ffd-46e576fd802e.png">

So please make mdRadio like this:
> <img width="111" alt="expect radio" src="https://cloud.githubusercontent.com/assets/15178410/23999905/b22dad0e-0a94-11e7-9bbc-5fed2f412e84.png">

PR #609, opened 2017-03-16
Enlarge mdRadio touchable area, like mdCheckbox (Do #608)

zcyzcy88:


marcosmoura:
Sounds great! Thank you for finding this! Merged!

PR #610, opened 2017-03-16
Update npm packages

d3radicated:


wopian:
😕

The only package this will affect is `webpack-merge`, as its the only one with a major version bump...

Morgul:
Additionally, this package uses yarn, so you will need to use [yarn upgrade](https://yarnpkg.com/en/docs/cli/upgrade), which will update the `yarn.lock` file.

marcosmoura:
This is already done in another branch. I will decline this PR, but I will give you the credits for this one, ok?

Thank you!

PR #620, opened 2017-03-21
replace undefined $checkbox-touch-size with $radio-touch-size

kpilard:
As an alternative, we could register a common value in variables.scss


marcosmoura:
I already fixed this issue by moving it to variables.scss.

Issue #593, opened 2017-03-07
 [mdInput] unwanted white line


zcyzcy88:
### Steps to reproduce
https://vuematerial.github.io/#/components/input
click and hold

### What is expected?
No white line

### What is actually happening?
<img width="63" alt="2017-03-07 20_15_00-components - input - vue material" src="https://cloud.githubusercontent.com/assets/15178410/23656101/3fa1678a-0373-11e7-8bb9-1eb14807e0a5.png">

FlyPointRed:
copy the file  **`\node_modules\vue-material\dist\vue-material.css`** ,
find this class **`.md-input-container .md-icon:not(.md-icon-delete):after`**,
delete this attribute  **`content: ""`**,
import this copy to replace your vue-material.css in index.html.
it can resolve this problem,and I didn't find it can cause any problem in other components.
I am looking forward to see the new version of this project.this project is so nice.

Samuell1:
Closing this issue as our focus is on the new 1.0.0 version.

PR #621, opened 2017-03-21
Remove empty content to remove empty line in icon (#593)

kpilard:


marcosmoura:
The is not the right fix for this issue. This "content" property should be there, but needs to be removed for other inputs. I will fix that.

PR #625, opened 2017-03-22
Fix subtotal not updating

d3radicated:
Subtotal on pagination is not updating when mdTotal value is updated (e.g. pagination data fetched from AJAX). This fixes it.


wxqee:
I have the same issue: http://codepen.io/wxqee/pen/peMOJJ

BTW, when should we get the version released to fix this issue?

hitautodestruct:
@marcosmoura Hey, I tried fixing this locally and compiling but alas no luck. Any chance to merge this pull request?

PR #646, opened 2017-03-28
Refactor Theme Engine

sirlancelot:
This is a first step toward supporting Server-side Rendering (SSR). Several other components still depend on `document` always being available, those will need to be fixed eventually. This PR will at least eliminate the theme engine as a bottleneck.

- Optimize parseStyle method to run a single RegExp replace
- Abstract DOM manipulation to check for VUE_ENV first
- Add theme name to `<md-theme>` element when it would render its own tag
- Recompute styles when a new set of selectors is added (via styles.push)
- Use computed property to find closest themed ancestor for mixin
- Only grab md-name attribute from an md-theme component (`<md-table-edit>` uses the md-name attribute for <input> tags)

### DIFF
--- a/src/core/components/mdTheme/dom.js
+++ b/src/core/components/mdTheme/dom.js
@@ -1,0 +1,38 @@
+export var changeHtmlMetaColor;
+export var createNewStyleElement;
+
+if (process.env.VUE_ENV !== 'server') {
+  changeHtmlMetaColor = (color, themeClass, previousClass) => {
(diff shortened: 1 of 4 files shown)


pablohpsilva:
I loved this pull request. Thank you!

sirlancelot:
Rebased to develop branch

marcosmoura:
WONDERFUL JOB!

PR #649, opened 2017-03-29
Add back click bubbling

sirlancelot:
When using the button component as part of a `<router-link tag="md-button">` the
event will not be attached to listen natively. Clicking on the button will not activate
the link.

```html
<!-- From the documentation -->
<router-link tag="md-button" to="/components/button" class="md-raised md-primary">Router Link</router-link>
```

Re-emitting the native event as a component event fixes the issue.


Morgul:
This is already being worked on, and click events will be added back to all relevant components. (see @marcosmoura's comment on [#541][541])

[541]: https://github.com/marcosmoura/vue-material/pull/541#issuecomment-287939408

Issue #631, opened 2017-03-23
[SCSS] md-hide-medium-and-up and md-hide-small both hidden at the same time


nahtnam:
### Steps to reproduce
1. Go to http://codepen.io/nahtnam/pen/ZeoJEv
2. Adjust the browser so it's 944px wide

Both the right menu button and the left sidenav icon are hidden at the same time making the menus inaccessible.
### Which browser?

I'm using Vue `2.2.4`, Vue Material `0.7.1`, `macOS Sierra`, and Chrome `56.0.2924.87`
### What is expected?
At least one of the buttons should be showing, either the small one or the medium and up one.
### What is actually happening?

Both buttons are hidden.
### Reproduction Link

http://codepen.io/nahtnam/pen/ZeoJEv

rascoop:
I think the problem is in the mixins.scss.
The **max-width** should be the breakpoint not the breakpoint - 16px or 17px

@mixin layout-xsmall {
  @media (**max-width: #{$breakpoint-xsmall}**) {
    @content;
  }
}

@mixin layout-small {
  @media (**max-width: #{$breakpoint-small - 16px}**) {
    @content;
  }
}

@mixin layout-medium {
  @media (**max-width: #{$breakpoint-medium - 16px}**) {
    @content;
  }
}

@mixin layout-large {
  @media (**max-width: #{$breakpoint-large - 17px}**) {
    @content;
  }
}

DavidBradbury:
Yeah, I'm having a similar issue with md-hide-small-and-up - That code looks ridiculous. 

`@mixin layout-small-and-up {
  @media (min-width: #{$breakpoint-small - 300px}) {
    @content;
  }
}`

Why is there hardcoded pixel values there?  Small and up would be -360 if you did want to hardcode a value... But it should be using the actual breakpoints to avoid unpredictable behavior around those points.

pablohpsilva:
Hello guys! How do you do?
The `16px` on those breakpoints have a reason:
> Desktop breakpoints are 16dp below the listed values to accommodate variations in browser chrome.

Let's talk about the breakpoints, shall we? Well, we have these breakpoints: 
```scss
$breakpoint-xsmall: 600px !default;
$breakpoint-small: 960px !default;
$breakpoint-medium: 1280px !default;
$breakpoint-large: 1920px !default;
```

Based on [Material Design Devices Screen sizes](https://material.io/devices/), the breakpoints, in my opinion, should be:
```scss
[...]

PR #668, opened 2017-04-06
Mixins: Removed arbitrary pixel values. Fixed “and-up” values. Added helper warnings. (#631) 

DavidBradbury:
Mixins included arbitrary offsets of the breakpoints. This was causing unpredictable behavior when some items would break at (for example) 600px, and others at 583px. (See #631)

"and up" values should be using the next breakpoint down instead of trying to math it, as the breakpoints may be changed by either the project or the user. (It should be noted that adding 1px is appropriate. Otherwise you get some items hidden, some not at exactly the breakpoints)

Helper warnings there for people who might try to "guess" a mixin without quite thinking what that would do.

### DIFF
--- a/src/core/stylesheets/mixins.scss
+++ b/src/core/stylesheets/mixins.scss
@@ -29,49 +29,61 @@
 }
 
 @mixin layout-small {
-  @media (max-width: #{$breakpoint-small - 16px}) {
+  @media (max-width: #{$breakpoint-small}) {
     @content;
   }
 }
 
 @mixin layout-medium {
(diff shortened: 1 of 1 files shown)


marcosmoura:
Those values are not arbitrary. They have a reason:
> Desktop breakpoints are 16dp below the listed values to accommodate variations in browser chrome.

DavidBradbury:
So this 'feature' seems to be causing issues when you're resizing the browser, creating odd breakpoints. How are the amounts calculated? Did you not wish to include the other useful portions, even if you don't think odd breakpoints are an issue?

PR #692, opened 2017-04-20
Dont overwrite currentSize of mdTablePagination

pschmidt88:
No matter what value you define as `md-size` option on `md-table-pagination`, it is overridden by the first value in the `md-page-options`.

It was not possible to define pagination options with `[10, 25, 50, 100]` and define the default start size as `25`.

### DIFF
--- a/src/components/mdTable/mdTablePagination.vue
+++ b/src/components/mdTable/mdTablePagination.vue
@@ -49,7 +49,7 @@
         subTotal: 0,
         totalItems: 0,
         currentPage: 1,
-        currentSize: this.mdSize
+        currentSize: parseInt(this.mdSize, 10)
       };
     },
     watch: {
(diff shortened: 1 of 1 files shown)


Samuell1:
@Morgul @pschmidt88 i tested it and doesnt work :/

pschmidt88:
I'll take a look on that when I've have some time.

pschmidt88:
@Samuell1 the last commit should fix this.

Samuell1:
@pschmidt88 doesnt work

pschmidt88:
How do you test it? I have it working in one of mine applications.

pschmidt88:
I think the problem is when you are passing the value directly as a string instead of binding it.
The latest commit should fix that, too.

Samuell1:
Yeah that is working

PR #708, opened 2017-04-26
Allow object for option value in select

Samuell1:


Samuell1:
I found one issue with this: In multiple select doesnt return all selected objects but only last selected

Morgul:
@Samuell1 If you haven't already, please file a bug for that, and reference this issue. (It'll just help us keep track of things.)

PR #536, opened 2017-02-20
Fix for bug that allowed disabled options in Select component to be selected

praneetloke:
Fix for #529. Updated docs with an example of disabled items in the Select component.

Issue #529, opened 2017-02-17
 [mdMenu] Disabled md-option can be selected 


cawa-93:
### What is expected?
In SELECT are some options with the attribute "disabled." These options change self opacity, but can still be selected and change the view

### What is actually happening?
When you click on an element that has an attribute "disabled" should not happen anything.

### Reproduction Link
https://jsfiddle.net/kozack/y7owLuz1/

marcosmoura:
Wow. This is a bad bug.

Samuell1:
Closing this issue as our focus is on the new 1.0.0 version.

PR #733, opened 2017-05-08
Fix for #529. Updated docs with an example of disabled items in the Select component.

praneetloke:
Follow-up for #536. Re-open PR from develop branch.

Note: This also contains a fix for the table pagination component.


praneetloke:
@marcosmoura here you go.

PR #773, opened 2017-05-24
Fix default theme not applying

d3radicated:


pablohpsilva:
You're awesome @d3radicated ! Thank you for the fix

PR #776, opened 2017-05-24
Add tabindex -1 to the input inside the switch

JeffBorges:
The tab passed twice by mdSwitch


pablohpsilva:
Thank you @JeffBorges

PR #778, opened 2017-05-25
Bunch of fixes

d3radicated:
- Allow md-checkbox to be toggled by clicking the label
- Add capability to edit chips in md-chips
- Fix error in md-select when not used with md-multiple

### DIFF
--- a/src/components/mdCheckbox/mdCheckbox.vue
+++ b/src/components/mdCheckbox/mdCheckbox.vue
@@ -5,7 +5,7 @@
       <md-ink-ripple :md-disabled="disabled" />
     </div>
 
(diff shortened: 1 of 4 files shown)


Samuell1:
Checkbox is toggled by label if have defined ID, its native functionality in html, doesnt needed to be added next click event..

d3radicated:
Oh, I see. But, in `md-checkbox`'s CSS, the cursor is pointer for the whole component which may be misleading that it doesn't toggle on the label if ID is not defined.

Samuell1:
@d3radicated maybe say that in docs will be better, then adding new method

pablohpsilva:
Thank you once again @d3radicated !!! You rock, dude!

PR #780, opened 2017-05-25
fix issue with mdSelect assuming multiple selection

tarciosaraiva:


tarciosaraiva:
@pablohpsilva here you go, on `develop` branch

PR #846, opened 2017-06-21
Small fixes

d3radicated:
- Fix md-icon-button not white on md-card-media-cover
- Allow custom md-ratio
- Fix cursor being a pointer when md-checkbox/md-radio is disabled
- Allow md-radio to be toggled even if id/name is not set
- Fix select all checkbox not updating when all md-table-row's are selected


pablohpsilva:
Thank you @d3radicated !! :D

PR #853, opened 2017-06-22
Fix #851

### LINKED ISSUE #851: Tag 0.7.3 breaks simple md-select
### Steps to reproduce
Go to the demo site : https://vuematerial.github.io/#/components/select
Click on one of the select element (the simple one, not the multiple selection ones)
Click on one of the value in the dropdown list

### Which browser?

At least Firefox and Chrome

### What is expected?
The value is changed on click

Nothing happens (nor the drop down list closes nor the value is changed)

### Reproduction Link

https://vuematerial.github.io/#/components/select
(it's also true for the version select in the header toolbar)

d3radicated:

### DIFF
--- a/src/components/mdList/mdListItem.js
+++ b/src/components/mdList/mdListItem.js
@@ -12,7 +12,7 @@ export default {
   },
   render(createElement, { children, data, props }) {
     const getItemComponent = () => {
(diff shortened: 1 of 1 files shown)


pablohpsilva:
Thank you for the quick fix @d3radicated :D

Issue #830, opened 2017-06-19
mdTablePagination mounted() warning


lucasfernog:
Vue development mode warning:

[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "mdPageOptions"

found in

---> MdTablePagination

`this.mdPageOptions = this.mdPageOptions || [10, 25, 50, 100]`

Fleeck:
Same issue here.

PR #854, opened 2017-06-22
Small fixes on mdChips, mdTablePagination (issue #830) and mdInputContainer buttons

lucasfernog:
Removed mdInputContainer clear/password button tabindex for better tab navigation in forms.

Debounce = 0 on mdChips input allows user to fastly type and add/remove chip (currently, the user has to wait or press enter repeatedly to add chip, since there is a verification for empty "currentChip" model value). Also, the user could remove a chip even with text on input due to the default debounce.


pablohpsilva:
That's cool! Thank you! :D

Issue #849, opened 2017-06-22
Defining and setting a custom theme is not working


petarpanicVIU:
### Steps to reproduce
Define a custom theme and set it afterwards globally with Vue.material.setCurrentTheme;

### Which browser?

Vue Material 0.7.2, Vue 2.3.4, every browser.

### What is expected?
The body element should have the theme set as a class.

### What is actually happening?
The body element is either empty or set to default theme

The last version that worked perfectly fine was 0.7.0.

pablohpsilva:
We'll look into it ASAP. Thank you @petarpanicVIU :D

PR #883, opened 2017-06-30
Theming fix - bug #849

petarpanicVIU:
The bug was submitted by myself and today I found some time to investigate the issue.

It seems there were two issues at hand. One was that the default theme class wasn't being set at all because the default class was already being set as an active class and with the refactored code you cannot set the already set theme. 
So I changed the default set theme in data object to be explicitly null and I'm setting a default theme on created event.

The other issue was that the custom theme wasn't being set, but it turned out it was being set, but the body class wasn't being applied. The reason was that the call to the function to set the body class had wrong parameter order. So I fixed the ordering and it's working correctly now.

### DIFF
--- a/src/core/components/mdTheme/index.js
+++ b/src/core/components/mdTheme/index.js
@@ -110,7 +110,7 @@ function injectStyle(style, spec, name, styleId) {
 export default function install(Vue) {
   Vue.material = new Vue({
     data: {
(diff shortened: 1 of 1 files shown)


pablohpsilva:
Thank you @petarpanicVIU  :D 
It makes sense. You rock!

Issue #887, opened 2017-07-03
Maxlength not enforced on md-dialogue-prompt


gabrielbonner:
I'm expecting to see a character limit enforced for prompts, like in the fourth example here -http://vuematerial.io/#/components/dialog

However in my codepen (with the exact same code from above) you'll notice that there is no character limit - https://codepen.io/anon/pen/dReRwd

Even if I add `:maxlength="prompt.maxlength"` to the element, still nothing.

Thoughts?  :)

zapping:
The property is `:md-input-maxlength` can you try with this one and check.

```
<md-dialog-prompt
    :md-title="prompt.title"
    :md-ok-text="prompt.ok"
    :md-cancel-text="prompt.cancel"
    :md-input-maxlength="prompt.maxlength"
    @open="onOpen"
    @close="onClose"
    v-model="prompt.value"
    ref="dialog5">
  </md-dialog-prompt>
```

gabrielbonner:
@zapping Thanks, I'll try it on Wednesday, we have a holiday tomorrow in the US.

gabrielbonner:
@zapping Yay, that worked, thanks.  Maybe updating the docs on the fourth example [here](http://vuematerial.io/#/components/dialog) would be in order so that other people don't have the same issue?

Thanks for your help.  👍

PR #892, opened 2017-07-05
Add missing properties to demo code #887

gabrielbonner:


pablohpsilva:
Thank you @gabrielbonner  :D 
We really let that slide.

PR #904, opened 2017-07-10
Fix syntax errors

ChalkPE:
[screenshot from 2017-07-10 12-09-37](http://vuematerial.io/#/components/tabs)


pablohpsilva:
Thank you @ChalkPE :D

Issue #686, opened 2017-04-15
 [mdList] md-list-expand inside a dialog box are open by default


arnoldmq:
When putting a list with expandable items inside a dialog box, the expandable items are open automatically, and some of those item in the expand element go under the main list items

Steps to reproduce
- create a dialog box,
-put list with expand items inside,
-make dialog box big enough that the list will fit in it with the expandable items open,
-open dialog box
-now that expandable items are open automatically 

Which browser?
chrome , ie

What is expected?
md-list-expand remain close when dialog is open

What is actually happening?
md-list-expand are open when dialog is open

Reproduction Link
http://codepen.io/anon/pen/dWobzK
(to see the issue make the result window in codepen bigger than it usually , if the md-dialog is small then the issue doesnt show, its mostly when the dialog box is bigger)

coppee:
Any clue ?

vizv:
Check `margin-bottom` of `md-list-expand`, it gives `0` instead of the `scrollHeight` of `md-list-expand`, and that's why it looks like already expanded on open.

I have a similar issue for the `margin-bottom` of `md-list-expand` may relate to your problem. A PR should come soon.

mlg-ipleic:
Anyone found solution to this, latest release is not working for me :/ still reproducible

vizv:
Well, PR #911 is a little bit different case. Could you guys check `active` state of `md-list-expand`. I think it might be the `active` state is somehow be set to `true`.

gywgithub:
I have the same problem.  
`md-liste-expand ` opens automatically
version: 0.7.5 
Code snippet: 
~~~

 <div id="app"> 
  <div>
    <md-button class="md-warn" @click.native="showCard">click</md-button>
    <md-card class="my-card" v-show="cardStatus">
      <md-list>
        <md-list-item>
          <md-icon>whatshot</md-icon>
          <span>News</span>

          <md-list-expand>
            <md-list>
              <md-list-item class="md-inset">World</md-list-item>
              <md-list-item class="md-inset">Americas</md-list-item>
              <md-list-item class="md-inset">Europe</md-list-item>
            </md-list>
          </md-list-expand>
        </md-list-item>
      </md-list>
    </md-card>
  </div>
</div>


<style>
.my-card {
 width: 200px;
}
</style>

<script>
Vue.use(VueMaterial)

var App = new Vue({
  el: '#app',
    data: {
      cardStatus: false
    },
    methods: { 
      showCard() {
[...]

PR #911, opened 2017-07-13
Fix incorrect padding for md-list-expand.

vizv:
Padding calculation should include `.md-list-expand` alone, not the whole parent `md-list-item`.

Try the fix with the following template:

```pug
<template lang="pug">
md-list-item
  .md-list-text-container
    span Line 1
    span Line 2
    span Line 3
    span Make sure the parent `.md-list-text-container` is high enough
    span And the div will be cropped if this patch haven't been applied

  md-list-expand
    md-list
      md-list-item.md-inset
        p foo
      md-list-item.md-inset
        p foo
</template>
```

May related to #686.

Signed-off-by: Wenxuan Zhao <viz@linux.com>

### DIFF
--- a/src/components/mdList/mdListItemExpand.vue
+++ b/src/components/mdList/mdListItemExpand.vue
@@ -60,7 +60,7 @@
       },
       calculatePadding() {
         window.requestAnimationFrame(() => {
-          this.height = -this.$el.scrollHeight + 'px';
+          this.height = -this.$refs.expand.scrollHeight + 'px';
 
           window.setTimeout(() => {
             this.transitionOff = false;


vizv:
Hold on, I should come up with a better way to fix this.

vizv:
This is good enough and ready for merge.

pablohpsilva:
You did came up with a better hay. Small and elegant. Thank you @vizv :D

mlg-ipleic:
I can't find that fix inside code when i install package with npm

belerofont:
Can you please verify this? because I don't see the code change when I installed cue-material with npm

vizv:
@belerofont Well, mine works. Could you provide code you have problem with (to reproduce)?

Issue #926, opened 2017-07-27
When opening a snackbar while it's already open, the snackbar doesn't close and then open


petarpanicVIU:
### Steps to reproduce
Try to click the open the snackbar button in the docs or just try opening the snackbar multiple times in the code.

### Which browser?

Every browser, every OS, every vue, the latest vue-material which is 0.7.4 at this point.

### What is expected?
Close the old snackbar and show the new one as that is the 'Material' way.

### What is actually happening?

The snackbar either stays open and just the message changes or it disappears and sometimes doesn't appear again.

pablohpsilva:
Thank you for registering this issue and fixing it on PR #928 LOL
You rock dude! :D

PR #928, opened 2017-07-27
Close snackbar before opening a new one - #926

petarpanicVIU:
When trying to show a snackbar when it's already shown,it was staying open or in some cases it disappeared, never to appear again. And trying to do a close and then open programmatically, it would always fail and would just hide the snackbar.

This adds a functionality that would hide the previous snackbar and open a new one, when you open a new one while the previous is still open.


pablohpsilva:
I think it looks good @petarpanicVIU 
Thank you!

PR #929, opened 2017-07-28
Both elements hidden when using hide-{size}-and-up / hide-{size-1}

Madsn:
Quick fix for issue with combinations of md-hide-{size n} / md-hide-{size n+1}-and-up, where both elements are hidden at the same time.

This occurs when the screen width matches the exact breakpoint.

https://codepen.io/Madsn/pen/jLbeEZ for samples

Related to #631 - or possibly closes it as an issue, though there is some discussion on an enhancement to the layouts going on.

### DIFF
--- a/src/core/stylesheets/mixins.scss
+++ b/src/core/stylesheets/mixins.scss
@@ -59,19 +59,19 @@
 }
 
 @mixin layout-small-and-up {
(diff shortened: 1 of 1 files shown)


pablohpsilva:
Thank you @Madsn 
I won't close issue #631 for now But I'll provide a better docs on that matter soon.
Thank you for your PR! We appreciate it.

PR #949, opened 2017-08-03
Added support for custom icons in md-tab with md-icon-src attribute

argarak:
This will allow for custom SVG/PNG icons to be used in `md-tab`. The icon source can be specified with the `md-icon-src` attribute and uses the existing `md-src` attribute in `md-tab`. I renamed it to `md-icon-src` so that it is obvious that it is an icon source and not anything else.

Example:
```
<md-tabs>
  <md-tab md-icon-src="/icons/icon.svg"></md-tab>
</md-tabs>
```

Works well with labels too.

### DIFF
--- a/src/components/mdTabs/mdTab.vue
+++ b/src/components/mdTabs/mdTab.vue
@@ -15,6 +15,7 @@
       mdLabel: [String, Number],
       mdIcon: String,
       mdIconset: String,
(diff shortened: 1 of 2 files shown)


pablohpsilva:
Cool! Simple and elegant. Thank you @argarak :D

PR #954, opened 2017-08-06
A really simple fix for avoiding double `change` trigger when clicking on md-radio's label

### LINKED ISSUE #953: `change` event triggers twice when clicking on a md-radio's label
### Steps to reproduce  
Click on a label from the ```md-radio``` component

### Which browser?
Google Chrome 59.0 and Firefox 47.0 with vue 2.3.3, vue-material 0.7.4

### What is expected?
Change event should trigger once

Change event triggers twice when clicking on label (but not when clicking directly on radio container)

### Reproduction Link
https://codepen.io/decksterr/pen/mMROyL  
Check the console for logs

**Note:** change event also triggers even if value of the related ```v-model``` didn´t actually change, although i am not sure if that is indeed desired behaviour

decksterr:
From what i think i understood, when clicking on a radio´s label the event is tied to the input itself and therefore bubbles up to the radio container, triggering the change event twice on the way.
This is, imo, undesired behaviour.
Adding ```.prevent``` on the event handler of the label (like in the checkbox component) will prevent default behaviour of tying to input and therefore only trigger change once.
https://vuejs.org/v2/guide/events.html#Event-Modifiers
fixes #953

### DIFF
--- a/src/components/mdRadio/mdRadio.vue
+++ b/src/components/mdRadio/mdRadio.vue
@@ -1,11 +1,11 @@
 <template>
   <div class="md-radio" :class="[themeClass, classes]">
-    <div class="md-radio-container" @click.stop.self="toggleCheck">
+    <div class="md-radio-container" @click.stop="toggleCheck">
       <input type="radio" :name="name" :id="id" :disabled="disabled" :value="value">
       <md-ink-ripple :md-disabled="disabled" />
     </div>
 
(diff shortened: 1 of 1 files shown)


decksterr:
quick and dirty fix is to not provide ids to radio elements to avoid tying the label to its respective input

decksterr:
quick and dirty fix is to not provide ids for radio elements

pablohpsilva:
Cool!! Thank you @decksterr 
Very simple and elegant fix.

Issue #896, opened 2017-07-06
header-item slot for mdTabs is not in latest release (0.7.4)


Nirazul:
### Steps to reproduce
The header-item slot for mdTabs is not in the latest release of `vue-material@0.7.4` as stated in the documentation.
According to this [PR](https://github.com/vuematerial/vue-material/pull/596) I thought that it would be in the next release. However, it's not...

### Which browser?
All browsers

### What is expected?
Ability to populate a slot for a custom mdTabs header.

### What is actually happening?
It's not included in the latest release, although the PR was already merged into develop.

argarak:
I took a look at the history of src/components/mdTabs/mdTabs.vue (in the `develop` branch) and it appears that commit https://github.com/vuematerial/vue-material/commit/500e5b5beac99421b2440b976918a30b404b424f has removed the header-item functionality. I'll submit a PR to put it back shortly.

PR #955, opened 2017-08-06
Fix header-item templating (issue #896) and tooltips on mdTab

argarak:
This fixes the `header-item` slot functionality in mdTabs as stated in the documentation. This was initially implemented in PR #596 but was removed in commit 500e5b5 (possibly merging an older branch).

I had to make a separate PR after my previous one as I had to move the element for `md-icon-src` functionality to be in `header-item`, if not there would be merging issues.

This PR also fixes tooltips on mdTab by making the `md-tooltip` element a child of the `tab-header` button element and not `md-tab-header-container`.

And finally, I also fixed a very slight shift in tooltip text as it moves. Here's an example of how this looks like:

text-shift-before
*The text moves a little.*

And here's an example of letter-spacing being applied to `md-tooltip` (after fix):

text-shift-after
*There is no longer any shift in the text.*

Feel free to change the letter-spacing if you think it is a little too much.

### DIFF
--- a/src/components/mdTabs/mdTabs.vue
+++ b/src/components/mdTabs/mdTabs.vue
@@ -21,10 +21,10 @@
                 <md-icon v-else-if="header.iconSrc" :md-src="header.iconSrc"></md-icon>
 
                 <span v-if="header.label">{{ header.label }}</span>
-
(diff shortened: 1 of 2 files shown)


pablohpsilva:
That's beautiful!
Awesome work @argarak you rock dude

PR #956, opened 2017-08-08
Add an option md-hint

bichikim:
It's md-hint like  Vuetify Text fields pros hint


argarak:
Thanks for contributing, @bichikim.

However, could you please work from the develop branch? The master branch is like the stable version of this repository and is only updated when there is a new release. So what I would recommend you do is to close this PR, save mdInputContainer.scss outside of the repository, reset `master` to before your commit, change branch to develop (you may also make another branch out of develop if you like) and overwrite mdInputContainer.scss from the one you saved locally. Finally, commit your changes.

You may also be able to use the `cherry-pick` command if you are confident in using `git` but this was the easiest way in my opinion.

I know that CONTRIBUTING.md does not make is very clear that you need to work in the develop branch to push PRs against the develop branch, I might change it for this reason.

pablohpsilva:
Something went wrong with your PR. It says you changed 141 files. Could you check it out please?

PR #957, opened 2017-08-08
State that contributors should work in the develop branch (modify CONTRIBUTING.md)

argarak:
This is to specifically state for contributors to make changes to their `develop` (or `develop`-derived) branch and not `master`.

(See #956)


pablohpsilva:
True, that's true. Thank you once again @argarak 
:D

PR #962, opened 2017-08-11
Fixed an issue with md-autocomplete and added some additional functionality to it

petarpanicVIU:
- Fixed lost focus on input when the menu appeared with the results
- Added max-res prop which can be added to limit the number of results shown in the list
- Show dropdown menu when input receives focus (not just when typing)
- Added a possibility to set max-chars to 0 and show the list (not fetched, but list) immediately on input focus, with input being empty.
- Fix mdTablePagination.vue error because of which it did not compile (linter issue with unused var)
- Added max-res in the docs


pablohpsilva:
Holly shit! Thank you so much @petarpanicVIU 
That's an amazing work! :D

Issue #895, opened 2017-07-06
 [mdAutocomplete] Issues with md-autocomplete


zapping:
There are some issues with the autocomplete component. Have added a pen here with the component. https://codepen.io/zapping/pen/GmVapv

### Steps to reproduce
1. When the component is initially loaded the label appears raised even though the v-model is not assigned a value. But when it receives a focus and then a blur, without typing a value, then label transitions back to just above the input correctly.

2. When you start typing a value the component looses focus, probably when the debounce timeout occurs, and you cannot continue typing rest of your contents to the component.

3. When you start typing the entire list/array appears in the dropdown without filtering according to the content being typed in the component.

4. The document for the component has a typo instead of the `md-autocomplete` tag it has the `md-input` tag.

5. Is there a way to show the dropdown list of the autocomplete when it receives focus? Both when it is clicked as well as when you tab through to the component. Maybe add a property for it or provide an event for that.

### Which browser?
Vue and vue-material is the latest one from the CDN. Tried on chrome, firefox and IE edge. The OS is window.

### What is actually happening?
The data that is fed to the component is a :list. I have not tried the :fetch option.

### Reproduction Link
https://codepen.io/zapping/pen/GmVapv

devistar:
Yes md-autocomplete component is bugs.

argarak:
I've recently created a new branch on my fork of vue-material [here](https://github.com/argarak/vue-material/tree/autocomplete-fixes) and I'll see what I can do to fix this. I'm currently working on the menu to appear at the bottom and fill the entire width of the input. Here's how this currently looks:

![screenshot of autocomplete](https://user-images.githubusercontent.com/13645600/29045180-8520f414-7bba-11e7-8270-e5a0cdbf2ef6.png)

There's still a small issue with the x-positioning and it seems to always be offset by 8px, however this is still a work in progress and the focusing problem has not been fixed yet. The focusing issue happens because the autocomplete menu is just the `mdMenu` element re-purposed, so when you click on a button or element with a menu, it automatically focuses on that menu. However, for auto-completion we obviously want to avoid this.

However, the problem with filtering can be solved by using the `filter-list` attribute in `md-autocomplete`. You have to assign it to a filter function, for instance:

```
exampleFilter: function(list, query) {
    return list.filter(function(item) {
        return item.label.indexOf(query) === -1 ? false : true;
    });
}
```

Please note that this filter has to be changed to suit your specific use-case. For example, the data that I have looks like this:

```
[...]

PR #969, opened 2017-08-14
Fix various issues with md-autocomplete (issue #895)

argarak:
After long await, the fixes are ready to be merged into `develop`. Here is a list of fixes:

- Input no longer blurred by `md-menu`
- Items can now be selected with the keyboard (arrow-keys and return)
- The maximum height of the menu can now be set to be either dependent on window height or the number of items in the list
- Menu stays put to the bottom of the input
- Menu is resized to width of trigger element (input)

I have also provided an example in the Input page of the documentation and have updated the API tables.

I would highly encourage for anyone to test this before merging as there may be issues I have not noticed.

To those with write access, **do not merge this pull request at any cost unless I specify in a comment to this PR that it is ready to be merged. This is because I need to test this after I solve the conflicts.**

### DIFF
--- a/docs/src/pages/components/Input.vue
+++ b/docs/src/pages/components/Input.vue
@@ -758,11 +758,11 @@
                   colorFilter: function(list, query) {
                     var arr = [];
             
-                    for(var i = 0; i < list.length; i++) {
+                    for(var i = 0; i &lt; list.length; i++) {
                       if(list[i].name.indexOf(query) !== -1)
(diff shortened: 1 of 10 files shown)


argarak:
I think we are good for merging now after my tests, however please do test on your end just in case I missed something.

pablohpsilva:
👏 👏 👏 👏 
Amazing work! Goddamn...
Thank you @argarak

PR #971, opened 2017-08-15
Fix issues while mdPageOptions is false

VdustR:
hide the pagination label and assign currentSize as mdSize while mdPageOptions is false.

The issue:
peek 2017-08-15 10-45-ori

Fixed:
peek 2017-08-15 10-45


pablohpsilva:
Thank you @VdustR 
Simple and neat. Amazing!

PR #987, opened 2017-08-18
Update and fix dependencies

jbjorge:
You already mention in the readme that you want vue-material to be standalone and therefore bundle vue with it.
Would it be feasible to trust vue to not release breaking changes in minor/hotfix-updates? If so, I would be glad to have this merged to save some bandwidth and processing power with avoiding bundling two versions of vue.
I understand the notion of locking the version you're running internally, so if you discard this PR, no feelings will be hurt.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,1 @@
+dist/docs/
--- a/package.json
+++ b/package.json
@@ -39,9 +39,7 @@
     "lint": "eslint --ext .js,.vue --fix src docs",
     "release": "bash build/release.sh"
   },
-  "dependencies": {
-    "vue": "^2.4.2"
-  },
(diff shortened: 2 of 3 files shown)


jbjorge:
Digging through the source, it turns out the library doesn't lock in the version, only the generation of docs do. DevDependencies need to be installed to generate docs anyway, so this can safely be merged without any breaking changes to users.
Do you want me to add a PR for separating out docs from the npm-module? (will shave off a whopping ~146 mb from the download)

jbjorge:
To sum up:
vue-material now...
* depends on Vue@^2.4.2 through devDependencies and peerDependencies instead of as a dependency
* is a ~3mb download from npm instead of 150mb

Issue #968, opened 2017-08-14
[mdStepper] stepHeader not clickable


githrdw:
### Steps to reproduce
1. Go to the stepper docs
2. The steps are not clickable (I think they are meant to be)

### Which browser?
Windows 10
Chrome, IE11, Edge, Firefox
Last Vue version 0.7.4

### What is expected?
When clicking on a step in the stepHeader the stepper have to jump to that step

### What is actually happening?
Nothing, not even an error

### Reproduction Link
[Stepper docs](http://vuematerial.io/#/components/stepper)

### Solution
When I change `@click="setActiveStep(step)"` into `@click.native="setActiveStep(step)"` it works.
@Morgul changed in [7daa66d](https://github.com/vuematerial/vue-material/commit/7daa66daa7b7cf0e117814f9f2f9f79f2c66c2c9) many `@click.native` to `@click` including the steppers'

pablohpsilva:
Hello @githrdw. How are you? :D 
I couldn't find any trace of the information that the steps are clickable on the [Material.io Documentation](https://material.io/guidelines/components/steppers.html#)

The steps are used to show a hint of what you're doing, but the flow is controlled by buttons.

I'm closing this issue for now, OK? If you find that it can be clickable, reopen this issue and let me know.

Thank you dude :D

githrdw:
@pablohpsilva Hi I'm fine, how about you?

You're right that the information is not in the documentation (yet?), but I'd rather look at the source.
In [mdStepper.vue](https://github.com/vuematerial/vue-material/blob/master/src/components/mdStepper/mdStep.vue) there is set an `@click` listener with the `setActiveStep`-function.
When I change `@click="setActiveStep()"` to `@click.native="setActiveStep(step)"` it works.
I think the maker of this stepper has thought about it but did not implement it?

githrdw:
And besides, it is described in the Material.io Documentation as ['Editable step'](https://material.io/guidelines/components/steppers.html#steppers-types-of-steps)

pablohpsilva:
I am great, thank you @githrdw :DD 

You're right. I totally missed it. We'll fix it ASAP

Samuell1:
Closing this issue as our focus is on the new 1.0.0 version.

PR #990, opened 2017-08-19
Fix clickable [mdStepper] header and added md-error

githrdw:
Hi @pablohpsilva, 
I think that I have found a solution for fixing the clickable header (in #968).
In addition to that, I compared the Material.io again and saw that vue-material doesn't support errors in the step.
So tried to fix that (with an md-error prop in the stepper) but I would not mind if you throw it away :P 
I'm still learning code so I hope you like it.


githrdw:
Screenshot of the email validation in mdStepper

stepper_validation

PR #992, opened 2017-08-21
Fixing bugs autocomplete

petarpanicVIU:
There was an issue with autocomplete when the input field was pre-filled before mounting the component. In that case, the input is still not visible or available, but the watcher is trying to update it. Simple null check prevents the errors from appearing.

Also another issue was that the list wasn't showing filtered if the value was pre-filled because there was a bug in checks for min-chars. It's now fixed.


pablohpsilva:
@petarpanicVIU , could you please check your PR again? 139 files to merge seems to be something wrong with this PR.

Thank you :)

petarpanicVIU:
@pablohpsilva Hey, I'm closing this PR and I've created a new one, which has only one file changed :)

PR #994, opened 2017-08-21
Enables md-raised class on md-button-toggle

JobJob:
Enables adding an "md-raised" class to "md-button-toggles" with
`<md-button-toggle class="md-raised md-primary">`

screenshot (see last line: "works with text too")

vue-material-toggle-buttons

I find it clearer to see the buttons, and it's opt in, so I don't think there's much downside.

Thanks for your work on this library btw!


pablohpsilva:
Thank you @JobJob ! :D

PR #1024, opened 2017-09-09
Has changed chat link from gitter to slack

gserrg:
Warning: readme.md in master already changed the link. There will be a conflict when  merged developer to maste


pablohpsilva:
That is true. Thank you for the warning @gserrg :D

Issue #414, opened 2017-01-28
 [mdCheckbox] Multiple md-checkbox, bound to the same Array


jgiovanni:
Is it possible to achieve multiple checkboxes, bound to the same array, like[ Vue's example ](https://vuejs.org/v2/guide/forms.html#Checkbox)currently? Or do we need to use the events? Can't seem to do this.
```
<input type="checkbox" id="jack" value="Jack" v-model="checkedNames">
<label for="jack">Jack</label>
<input type="checkbox" id="john" value="John" v-model="checkedNames">
<label for="john">John</label>
<input type="checkbox" id="mike" value="Mike" v-model="checkedNames">
<label for="mike">Mike</label>
<br>
<span>Checked names: {{ checkedNames }}</span>
```
```
new Vue({
  el: '...',
  data: {
    checkedNames: []
  }
})
```

marcosmoura:
Not right now, but this is a great improvement. I'm going to do it.

capttrousers:
Or an object, something like 
```
checkedNames {
"John" : false,
"Jim" : true,
"Jill" : false,
"Jack" : true
}
```

Okipa:
How would you manage to currently handle the checked checkbox values ?  
I tried to use the @change event but in case of a page by page survey (on page for a question), I have to collect the answers and to reload them if we go back on a question page. I am kind of stuck here ...

giovanniorigins:
@Okipa so far, I've managed to find a workaround by manually building the checkboxes like this
```html
<template v-for="(g, $index) in $root.genresList">
                       <!-- What we want to use... -->
                        <!--<md-checkbox :value="g.id" class="">{{g.name}}</md-checkbox>-->

                    <!-- What we have to use... -->
                        <div class="md-checkbox md-theme-default" :class="{ 'md-checked': contains('selectedGenres', g.id) }">
                            <div class="md-checkbox-container">
                                <input :id="'genreCheckA'+$index" type="checkbox" v-model="selectedGenres" :value="g.id">
                            </div>
                            <label class="md-checkbox-label" :for="'genreCheckA'+$index">{{g.name}}</label>
                        </div>
                    </template>
```
```javascript
data(){
  return {
    selectedGenres: [],
  }
},
methods: {
  contains(arr, string){
    return _.contains(this[arr], string);
  },
}
```

This way it has the functionality of regular vue checkboxes with vuematerial design

Okipa:
[...]

PR #1029, opened 2017-09-13
#414 Checkbox accept Array as v-model

jspies:
This adds the ability to use an Array in the v-model for MdCheckbox. It provides the same functionality as Vue's [basic input](https://vuejs.org/v2/guide/forms.html#Checkbox).

This is a requested feature documented here: https://github.com/vuematerial/vue-material/issues/414


pablohpsilva:
Nice! Thank you @jspies :D

PR #1078, opened 2017-10-10
md-select menu same width 

petarpanicVIU:
md-select's md-menu component was not using the full width of the md-select.

Now, there is a new prop that can be passed to md-select which is then passed to the corresponding md-menu and then allows the full width dropdown.


marcosmoura:
I'm sorry for taking so long to answer this. I'm focused on the new version.
Can you change this?

src/components/mdSelect/mdSelect.vue
```diff
@@ -65,7 +66,8 @@
       disabled: Boolean,
       placeholder: String,
       mdMenuClass: String,
-      mdMenuOptions: Object
+      mdMenuOptions: Object,
+      menuFullWidth: Boolean
```
marcosmoura: add the md prefix for this prop

petarpanicVIU:
@marcosmoura Fixed it :)

PR #1147, opened 2017-11-16
fix(MdTable): fix MdTable.hasValue not synced issue

### LINKED ISSUE #1146: Table selection doesn't work if the table model is empty initially
### Steps to reproduce

Try to select a row in the provided codepen. The data is loaded asynchronously after a 200ms delay to simulate an AJAX call.

### Which browser?

All

### What is expected?

Table rows should be selectable

Table rows can't be selected, the "default" cursor is shown instead of "pointer"

As a workaround you can add v-if to the table, so the initialization will be done after the data is retrieved, which isn't ideal, because the table header isn't visible also (see https://codepen.io/danyalex/pen/qVPrEN)

### Reproduction Link
https://codepen.io/danyalex/pen/WXZRZd

VdustR:
Related to #1146

fix #1146

After table values changed, the `MdTable.hasValue` was not updated.

* Adding a watcher to make `MdTable.hasValue` follow the computed `hasValue`.
* Declaring all `MdTable` properties in `data()` instead of some declarations in `provide()`.

### DIFF
--- a/src/components/MdTable/MdTable.vue
+++ b/src/components/MdTable/MdTable.vue
@@ -103,21 +103,31 @@
         }
       }
     },
-    data: () => ({
-      fixedHeaderPadding: 0,
-      hasContentScroll: false,
-      MdTable: {
-        items: {},
-        sort: null,
-        sortOrder: null,
-        singleSelection: null,
-        selectedItems: {},
-        selectable: {},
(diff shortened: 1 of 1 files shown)


marcosmoura:
Sounds good!

PR #1156, opened 2017-11-17
fix(MdInput) typo in font size

emman27:
As mentioned in the title, this appears to be a typo where the font-size of a text area is inconsistently set at 1px instead of 16px. I've attached a gif marking this on the official docs site, though seriously at 1px you can't read what I've typed anyway
vuematerial

### DIFF
--- a/src/components/MdField/MdField.vue
+++ b/src/components/MdField/MdField.vue
@@ -184,7 +184,7 @@
       transition: $md-transition-stand;
       transition-property: font-size, padding-top, color;
       font-family: inherit;
-      font-size: 1px;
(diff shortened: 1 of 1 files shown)


marcosmoura:
Which browser?

emman27:
That gif is taken on Chrome 62.

akomarovsky:
The same thing happened to me. Seems like when Grammarly is on then the bug occurs

emman27:
Oh wow you're right there I didn't actually consider that when I took this. Just tried with it off and it works fine. So... Is the bug caused by vuematerial or grammerly?

Issue #1250, opened 2017-12-02
App Layout height - beta 1.0.6


C0d1ngJammer:
### Steps to reproduce

In my Layout I added the default App Layout "Reveal" [https://vuematerial.io/components/app](https://vuematerial.io/components/app)

This is what my layout looks like:

    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700,400italic|Material+Icons">

    <link rel="stylesheet" href="https://unpkg.com/vue-material@beta/dist/theme/default.css">

`<html>
<head>
    <meta charset="utf-8">
    <meta content="width=device-width,initial-scale=1,minimal-ui" name="viewport">

    <style lang="scss" scoped>
        /*.md-app {
            /*max-height: 400px;
            height: 100%;
        }*/
    </style>
</head>
<body>

    <div  id="app">
        <div class="page-container">
            <md-app md-mode="reveal">
                <md-app-toolbar class="md-primary">
                    <md-button class="md-icon-button" v-on:click="menuVisible = !menuVisible">
                        <md-icon>menu</md-icon>
                    </md-button>
                    <span class="md-title">My Title</span>
                </md-app-toolbar>

                <md-app-drawer :md-active.sync="menuVisible">
                    <md-toolbar class="md-transparent" md-elevation="0">Navigation</md-toolbar>

                    <md-list>
                        <md-list-item>
                            <md-icon>move_to_inbox</md-icon>
                            <span class="md-list-item-text">Inbox</span>
                        </md-list-item>

                        <md-list-item>
                            <md-icon>send</md-icon>
                            <span class="md-list-item-text">Sent Mail</span>
                        </md-list-item>

                        <md-list-item>
                            <md-icon>delete</md-icon>
                            <span class="md-list-item-text">Trash</span>
                        </md-list-item>

                        <md-list-item>
                            <md-icon>error</md-icon>
                            <span class="md-list-item-text">Spam</span>
                        </md-list-item>
                    </md-list>
                </md-app-drawer>

                <md-app-content>
                    <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. Error quibusdam, non molestias et! Earum magnam, similique, quo recusandae placeat dicta asperiores modi sint ea repudiandae maxime? Quae non explicabo, neque.</p>
[...]

Issue #1222, opened 2017-11-27
How to set full height in md-app


Isaiasdd:
How to set a full height in md-app?

![full_height](https://user-images.githubusercontent.com/5838676/33283174-2de23e62-d392-11e7-852c-113520867295.PNG)

emman27:
You could try
```
<md-app style="min-height: 100vh;">
  <whatever-you-need />
</md-app>
```
(or use a CSS class)

Isaiasdd:
@emman27 Thank you, it's working!

czy1996:
I tried adding class or style as such,
```
<md-app :class="appWrapper" class="app-wrapper">
```
```
.app-wrapper {
    min-height: 100vh;
  }
```
it didnt help, and html rendered didnt have the class added
but if i add `app-wrapper` in the chrome dev tool, the height changed
adding inline css didnt help too
my vue version is 2.5.2 and vue material is 1.0.0-beta-7

Samuell1:
@czy1996 for beta-7 you need to add, in next version is allowed to add classes on **md-app** component

```
.md-app {
   height: 100vh;
}
```

HappyBird233:
@Samuell1  
it complete slove my problem.  but i don't konw how come that the fixed class on md-app didn't work when i used  .md-app{ min-height: 100vh; }
sorry my english is poor

jsantoru:
@marcosmoura @Samuell1 @HappyBird233 i might be missing something... is there a sandbox example of this working? when i try to set the full height it doesn't seem to work: https://codesandbox.io/s/6yv4myjynr

i am just getting the outer browser scroll bar and losing the content only scrolling. any help is appreciated. thanks!

Samuell1:
@CreatixEA https://github.com/vuematerial/vue-material/issues/1222#issuecomment-361159516

PR #1254, opened 2017-12-03
feat(MdApp): add full height class

Samuell1:
https://github.com/vuematerial/vue-material/issues/1222, https://github.com/vuematerial/vue-material/issues/1250

### DIFF
--- a/docs/app/pages/Components/App/App.vue
+++ b/docs/app/pages/Components/App/App.vue
@@ -83,6 +83,7 @@
         <p>{{ $t('pages.app.apiApp.propsDesc') }}:</p>
 
         <api-table :headings="api.props.headings" :props="api.props.props" slot="props" />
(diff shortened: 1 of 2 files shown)


VdustR:
I think give it `100%` and let them decide the height from outer is more flexible in my opinion.

In some case someone might stick an ad on top or the bottom, or spit the screen up and down for another subpage.

marcosmoura:
I like both proposals. I think this should only come when we have the boilerplates.

Samuell1:
It would be better to add prop like `md-height` and default will be `100vh` ? @VdustR @marcosmoura

VdustR:
How about custom classes and style allowed?

https://github.com/VdustR/vue-material/commit/2a9fa0e2adf0560c13c8fde3d7dba284b0a8f933

marcosmoura:
@VdustR I liked it!

PR #1281, opened 2017-12-08
feat(MdChips): feedback for duplicated chip

### LINKED ISSUE #1212: [MdChips] feedback for duplicated chip input value
`MdChips` reject duplicated value.

It just do nothing while trying to insert a duplicated value.

There should be a feedback let users know it's duplicated, like: shake the duplicated chip or some else anime effects.

VdustR:
* New props `md-always-check-duplicated` would check if there is duplicated chip on input value change if set it as `true`.
* Class `md-duplicated` would be added on the duplicated chip while it was detected.
* Class `md-duplicated` has the same style with `md-accent` as default.
* You could customize `md-duplicated` style via `deep selector` if you are using scoped CSS.

peek 2017-12-08 13-41

fix #1212

### DIFF
--- a/docs/app/pages/Components/Chips/Chips.vue
+++ b/docs/app/pages/Components/Chips/Chips.vue
@@ -2,6 +2,7 @@
 <example src="./examples/Static.vue" />
 <example src="./examples/Editable.vue" />
 <example src="./examples/ChipCustomTemplate.vue" />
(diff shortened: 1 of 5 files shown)


marcosmoura:
THIS IS SO COOL!!! Awesome job!

PR #1285, opened 2017-12-08
(Fix #1261) Don't emit input directly in MdInput and MdTextarea

### LINKED ISSUE #1261: Multiple input events on MdInput
**Edit**: See [this](https://github.com/vuematerial/vue-material/issues/1261#issuecomment-349463097) comment for what I believe fixes this issue.

### Steps to reproduce

Use vue-dev-tools on the vue-material docs. (I can't seem to get the dev tools to activate on vuematerial.io, so I'm using a local copy a la `npm run dev`.)

Watch the events and type in an input, textarea or change a select.

### Which browser?

Latest Vue, git dev version of vue-material, latest stable Chrome

### What is expected?

I would expect to only receive one input event. In the case of the select, I would expect to receive one event with the value of the selection.

You'll see two or three events for each key stroke. In case of the select, you'll get one input event with the value of the selection and another with the inner text of the selection.

### Investigation

In the case of MdInput, there's three sources that these events come from. First in `listeners()` in `MdInput.vue`. 

In `MdFieldMixin.js` the `set(value)` for the computed `model`, and the watcher for `value` both set `localValue` which causes an emit.

I don't know enough about how MdInput is supposed to work to fix it. I did notice that `set(value)` for `model` checks if `value` is an `InputEvent`. In all my testing the only thing `value` ever is is a `String`.

korylprince:
Fixes #1261.

Notes: use `Object.assign` to deep copy `this.$listeners`. `delete` `input` key as opposed to setting to `undefined` which causes error.

### DIFF
--- a/src/components/MdField/MdTextarea/MdTextarea.vue
+++ b/src/components/MdField/MdTextarea/MdTextarea.vue
@@ -48,10 +48,9 @@
     },
     computed: {
       listeners () {
-        return {
-          ...this.$listeners,
-          input: event => this.$emit('input', event.target.value)
-        }
+        var l = Object.assign({}, this.$listeners)
+        delete l.input
+        return l
       },
       textareaStyles () {
         return {
--- a/src/components/MdField/MdTextarea/MdTextarea.vue
+++ b/src/components/MdField/MdTextarea/MdTextarea.vue
@@ -48,7 +48,7 @@
     },
     computed: {
       listeners () {
-        var l = Object.assign({}, this.$listeners)
+        var l =  {...this.$listeners}
         delete l.input
         return l
       },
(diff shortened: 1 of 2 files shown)


src/components/MdField/MdTextarea/MdTextarea.vue
```diff
@@ -48,10 +48,9 @@
     },
     computed: {
       listeners () {
-        return {
-          ...this.$listeners,
-          input: event => this.$emit('input', event.target.value)
-        }
+        var l = Object.assign({}, this.$listeners)
```
marcosmoura: `Object.assign` is not available for all browsers.

korylprince:
Sorry! I thought `Object.assign` was transformed by babel. I've changed it over to the spread operator which is transformed by your babel config.

Issue #1284, opened 2017-12-08
[MdDialog] title not support direction: RTL


omriLugasi:
md-dialog component not support `RTL` direction

### Which browser?
Chrome, Explorer, FF

### What is expected?

md-dialog should support rtl option with directive of `rtl-dialog`.

Should I create a pull request?

omriLugasi:
@Samuell1  Should i help you with `RTL ` issues through pull request?

Samuell1:
@omriLugasi there will be good to have global RTL support for all components (https://github.com/vuematerial/vue-material/issues/80)

Samuell1:
Closing this issue, please point any questions about RTL in issue #80

Issue #80, opened 2016-11-16
how about RTL support ?


alphaelf:
is possible to make components RTL by passing a variable?

marcosmoura:
If you change in the HTML tag to ```<html dir="rtl">``` you will see that some components will work as expected. But some other will not. Like Menus, for example. This will be implemented after release v0.5.0. Thanks!

alphaelf:
i'll be waiting, tnx

alphaelf:
i use [rtlcss.com](http://rtlcss.com/) to RTL main `vue-material.css` and it work fine.

MahdiMajidzadeh:
any good news yet?

VSG24:
Anything on this?

MahdiMajidzadeh:
@alphaelf, how about js component ??

vad11:
Please add RTL switch to theme definition.

PR #1312, opened 2017-12-16
Support RTL language issues: 80 , 1284

omriLugasi:
Add support RTL language in the components below:

1.md-bottom-bar-item
2.md-card-actions
3.md-auto-complete
4.md-datepicker
5.md-dialog-alert
6.md-dialog-confirm
7.md-dialog-prompt
8.md-checkbox
9.md-chip

This commit will be "Stage One" in supporting rtl languages;
Next pull request will support rtl languages in the components below:

1. All input components
2. List
3. Menu
4. Snackbar
5. Tabs
6. Tooltip

this pull request fix part of the problem at issue number : [80](https://github.com/vuematerial/vue-material/issues/80), and fix issue number [1284](https://github.com/vuematerial/vue-material/issues/1284)


omriLugasi:
@Samuell1  The is an issue with the travis 

```
src/components/MdField/MdField.test.js

  ● Test suite failed to run

    Vue packages version mismatch:  

    - vue@2.5.11

    - vue-template-compiler@2.5.9

    

    This may cause things to work incorrectly. Make sure to use the same version for both.

    If you are using vue-loader@>=10.0, simply update vue-template-compiler.

    If you are using vue-loader@<10.0 or vueify, re-installing vue-loader/vueify should bump vue-template-compiler to the latest.
```

>Please update  vue-template-compiler version to 2.5.11

Samuell1:
@marcosmoura can you look at that dependency?

@omriLugasi i think better will be some global variable to enable RTL, then adding prop to all components, what you think?

VdustR:
fixed in #1306 

You could install `vue@2.5.9` manually temporarily.

omriLugasi:
@Samuell1 Sure i can do it;
My idea was to let the user the option to write application with 2 options `RTL` **and** `LTR`, but i don't care to change it to global variable.

> I'll change the flow to global variable :)

omriLugasi:
@Samuell1 

>Ok 👍  :)  Global configuration will be add by passing an object options to the vueMaterial Instace 


```
const options = {
  mdRtl: true
}

Vue.use(VueMaterial , options);
```

marcosmoura:
I don't think it's a good idea to have an option to configure the content direction.
I think the best and clever way to do it is to hook the styles based on [dir="rtl"] html attribute. It will be safer and complaint with HTML5 specs.

marcosmoura:
I think that a PR per component would be better. It will be really hard to track which components have been upgraded.

marcosmoura:
c.c @omriLugasi

omriLugasi:
@marcosmoura 
@Samuell1 

Did you mean something like:

```
<md-dialog-alert
      :md-active.sync="second"
      md-title="Post created!"
      md-content="Your post <strong>Material Design is awesome</strong> has been created."
      dir="rtl"
 />

```

And in the md-dialog component under style section to do :

```
.md-dialog{
  ...
  &[dir="rtl"]{
     text-align: right;
  }
}
```

omriLugasi:
**Still waiting for your answer....**

@marcosmoura 

@Samuell1

marcosmoura:
No. dir="rtl" inside the HTML tag:
```
<html dir="rtl"> ... </html>
```

marcosmoura:
@omriLugasi Any news on that? :)

omriLugasi:
@marcosmoura 

There is an issue with this rule

Avoid qualifying attribute selectors with an element
`html[dir='rtl']`

in the codeclimate....

Anyway I committed our first support in the md-filed  `icons and labels`.

Please tell if you like it, and i can keep going with this line of structure...

Samuell1:
I think this PR should be closed and opened new after components are more stable. And there are many commits that only removing, moving code and that can make many conflicts in codebase.

Maybe RTL support should be in separate files (Like `rtl.scss` in every component folder) or as package?  What you think? @marcosmoura @omriLugasi

PR #1293, opened 2017-12-11
fix(MdSelect): make options in disabled groups un-selectable and fix checkboxes

VdustR:
* fix(MdSelect): make options in disabled groups un-selectable
    fix #1286
* fix(MdSelect): make multiple selection checkboxes effective
    before:
        peek 2017-12-11 10-56
    after:
        peek 2017-12-11 10-55

PR #1318, opened 2017-12-19
Fix conflict between MDOption and MDCheckbox onchange events

mikatalk:
Currently there is a conflict in the Multiple Selection component.
Clicking on the main button surface area behaves as expected. 
However, clicking inside the checkbox seems to be toggled twice by both MDCheckbox then MDOption.

The fix removes pointer events from checkbox in the context of a select/option.

I'm not yet too familiar with the project style, so this may not be the proper way of adding a new class. Please let me know and I can address it.

### DIFF
--- a/src/components/MdField/MdSelect/MdOption.vue
+++ b/src/components/MdField/MdSelect/MdOption.vue
@@ -1,6 +1,6 @@
 <template>
   <md-menu-item :class="optionClasses" :disabled="isDisabled" @click="setSelection">
-    <md-checkbox class="md-primary" v-model="isChecked" v-if="MdSelect.multiple" />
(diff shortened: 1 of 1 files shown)


VdustR:
Seems duplicated with #1293

mikatalk:
@VdustR Yeah it seemed to be the same problem, I only opened the PR to point to a different solution. 
There is no case I can think of where we would want the checkbox to maintain its pointer events, therefore i thought CSS would be an easy fix. 
Is there any concern with this approach?

marcosmoura:
Yes. I see a problem with the ripple feedback for mobile devices. Using pointer events will prevent the feedback radius on checkboxes.

PR #1338, opened 2017-12-25
 fix(MdAutocomplete): options with objects

### LINKED ISSUE #1243: [MdAutocomplete] options with objects.
As #1215, I fixed by #1218 to make `searchTerm` sync with `v-model`, but that also make it crash while the options are objects. It's could reproduce in documentation example **Custom Template** - **With highlight text** with 6f51eb91338a0fa845a99b2a65f1c76f8d549196 (and it's a fatal error now in beta 6). If I want to make #1215 support object, I should know what to show as `searchTerm`. I think it's not easy because `MdAutocomplete` get the plain text via [`Node.textContent`](https://github.com/vuematerial/vue-material/blob/cb795934daab18df84b22e11c3be8675209bf6a1/src/components/MdAutocomplete/MdAutocomplete.vue#L206).

Another issue is that `MdIcon` cannot be used in `md-autocomplete-item` slots or the `searchTerm` will show the text in `MdIcon`.

As #1241, I think users prefer that the result is the thing shown on the `searchTerm` not the thing they select via autocomplete. It's good if I could still get the object while I type the `searchTerm` and match a statement.

Any idea?

VdustR:
BREAKING CHANGE: Selected object only be passed via `@md-selected`, `v-model` is always bound with `searchTerm`.

* `v-model` type is changed to `[String, Number]` and always be bound with `searchTerm`.
* `@md-selected` pass selected item. This could be used to get the selected object. This event would be emitted on next tick after `v-model` changed because the case that clear selected item while `searchTerm` changed considered.
* new props `:md-get-plain-text` aim to ignore content with iconic font added.

fix #1243

### DIFF
--- a/docs/app/pages/Components/Autocomplete/Autocomplete.vue
+++ b/docs/app/pages/Components/Autocomplete/Autocomplete.vue
@@ -149,8 +149,8 @@
           props: [
             {
               name: 'v-model',
-              type: 'String|Number|Boolean|Array',
               description: 'The model variable to bind the autocomplete value',
+              type: 'String|Number',
               defaults: 'null'
             },
             {
--- a/docs/app/pages/Components/Autocomplete/Autocomplete.vue
+++ b/docs/app/pages/Components/Autocomplete/Autocomplete.vue
@@ -3,6 +3,7 @@
 <example src="./examples/AutocompleteSearch.vue" />
 <example src="./examples/AutocompleteBox.vue" />
 <example src="./examples/AutocompleteTemplate.vue" />
+<example src="./examples/AutocompleteObjectText.vue" />
 <example src="./examples/AutocompleteAsync.vue" />
 
 <template>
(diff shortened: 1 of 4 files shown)


marcosmoura:
@VdustR Is there any way that you think of about not turning this into a breaking change, but keep the object support for autocomplete?

VdustR:
While `v-model` bound with `searchTerm`, it's already a breaking change becuase searchTerm don't support objects in my opinion.

lokkilok:
Ran into the same problem, and had a look at this Pull Request. I wonder if it could be better to fix this such that `searchTerm` would be a computed attribute such that if a function like getPlainText is given that function is used to return a String, and otherwise the value as such (or perhaps cast as a String) is returned. Just a thought.... haven't tried that myself yet.

lokkilok:
Played around a bit with that idea and came up with a slightly different solution, using just a field name on the item that should return a String. Probably also a breaking change, as parent components that want to update with an Object when an Object is selected will have to listen to `md-selected`. I'm pretty new to Vue and vue-material, so I've no idea if the following path (against the dev branch version) is any better...
`
@@ -73,11 +73,12 @@
       mdInputName: String,
       mdInputId: String,
       mdInputMaxlength: [String, Number],
-      mdInputPlaceholder: [String, Number]
+      mdInputPlaceholder: [String, Number],
+      mdValueTextField: String
     },
     data () {
       return {
-        searchTerm: this.value,
+        searchTerm: (this.mdValueTextField && this.value) ? this.value[this.mdValueTextField] : this.value,
         showMenu: false,
         triggerPopover: false,
         isPromisePending: false,
@@ -136,7 +137,7 @@
       },
 
       value (val) {
-        this.searchTerm = val
+        this.searchTerm = (this.mdValueTextField && val) ? val[this.mdValueTextField] || val : val
       }
     },
     methods: {
@@ -205,7 +206,7 @@
         this.$emit('md-closed')
       },
       selectItem (item, $event) {
-        const content = $event.target.textContent.trim()
+        const content = this.mdValueTextField ? item[this.mdValueTextField] : $event.target.textContent.trim()
 
         this.searchTerm = content
         this.$emit('input', item)
`

thomasklein:
Hi! Any updates on this PR?

LaravDev:
Will this PR merge? any updates?

PR #1346, opened 2017-12-26
fix(MdSelect): fix dense

### LINKED ISSUE #1332: md-dense not working on <md-select>
### Steps to reproduce
Go to https://vuematerial.io/components/select/ at the "Dense" section and open a select-box (or in the provided CodePen)

### Which browser?
Chrome at least

### What is expected?

With the md-dense prop on the md-select component, items in list should be smaller.

It's not the case in the doc and on the provided CodePen.
md-dense class is not applied

VdustR:
make dense props available

fix #1332


marcosmoura:
Sounds Good!

Issue #1263, opened 2017-12-05
md-menu "Unhandled promise rejection" within IE 11 (edge mode)


tachiaus:
**Steps to reproduce**

Have an md-menu anywhere on page, when clicking to open, the menu won't open and the console gives an error, works in other browsers

**Which browser?**

IE 11 running in edge mode

VueMaterial 1.0.0-beta-6

**What is expected?**
Menu should open

**What is actually happening?**
Menu items don't display, console shows error 
`Unhandled promise rejection TypeError: Unable to get property 'toLowerCase' of undefined or null reference`

`   at _callee2$ (file:///C:/Users/benma/Desktop/vuematerial/my-project/dist/testmenu.js:22201:19)`

Error appears in this section of code
`
createPopper: function () {
 var _ref2 = _asyncToGenerator( /*#__PURE__*/regeneratorRuntime.mark(function _callee2() {
var options;
return regeneratorRuntime.wrap(function _callee2$(_context2) {
 while (1) {
switch (_context2.prev = _context2.next) {
case 0:
 if (this.mdSettings) {
 options = (0, _deepmerge2.default)(this.getPopperOptions(), this.mdSettings);
 if (this.$el.constructor.name.toLowerCase() !== 'comment') {
                    this.popperInstance = new _popper2.default(this.originalParentEl, this.$el, options);
                  }
                }
              case 1:
              case 'end':
                return _context2.stop();
            }
          }
        }, _callee2, this);
      }));
    }),
`
Specifically at

`  if (this.$el.constructor.name.toLowerCase() !== 'comment') {
                    this.popperInstance = new _popper2.default(this.originalParentEl, this.$el, options);
                  }`

**Reproduction Link**

```
<template>
  <div class="page-container">
          <md-menu md-size="medium" md-align-trigger>
            <md-button md-menu-trigger>Add task</md-button>
          <md-menu-content>
            <md-menu-item>Add incoming (quick)</md-menu-item>
            <md-menu-item>Add incoming (new tab)</md-menu-item>
            <md-menu-item>Add out going</md-menu-item>
          </md-menu-content>
        </md-menu>
  </div>
</template>
```

tachiaus:
This issue is also present in beta7

Samuell1:
Its weird because https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase is supported in IE

VdustR:
[...]

PR #1355, opened 2017-12-28
fix(MdPopover): constructor name

Samuell1:
Closes https://github.com/vuematerial/vue-material/issues/1263

### DIFF
--- a/src/components/MdPopover/MdPopover.vue
+++ b/src/components/MdPopover/MdPopover.vue
@@ -87,7 +87,7 @@
           this.createPopper()
         }
       },
-      getObjectName (obj) {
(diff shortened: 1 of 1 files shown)


VdustR:
Search `.constructor.name` over `src` folder. You could find there are other codes use this method determine constructor. Would you deal them all?

src/components/MdPopover/MdPopover.vue
```diff
@@ -87,11 +87,16 @@
           this.createPopper()
         }
       },
+      getObjectName (obj) {
```
VdustR: Shouldn't it be called `getConstructorName`?

PR #1358, opened 2017-12-29
 feat(MdTable): reactive selection

### LINKED ISSUE #1292: [MdTable] pre select items with selection
Is it possible to programmatically select items in the table?
I can't find the option in the api as well as when looking at the code of the table component.
Would be handy to have property for it for example:
``
<md-table>
            <md-table-row slot="md-table-row" slot-scope="{ item }" md-selectable="multiple" md-auto-select :md-is-selected="true">
``
If it's not possible currently, is my only option is to fork and add the option?

Thanks

### LINKED ISSUE #1348: [MdTable] Table Select-All Issue
### Steps to reproduce

Create table with elements that can be sorted by id and other types, and sort them in a way that breaks the linearity of the id numbers. Once the select all checkbox is clicked, only some are selected.
Edit - There seems to be a litany of issues with select all, the same issue also appears and is reproducable in the codepen below with many other cases,
- By searching for specific elements and selecting all, not all are selected (e.g. search for 't' and try to select all elements https://i.imgur.com/YHyIUlS.png
- Exit out of that search and again try to select all on all elements, all of a sudden a few elements are not selected when clicking the select all. https://i.imgur.com/3ELZgb2.png

### Which browser?

Vue-Material@^1.0.0-beta-7
Vue@^2.5.0
Electron@1.7.10

### What is expected?

It should select all elements regardless of id order

I believe it is using the id's to decide which elements to select, and when it encounters out of order id's it goes off.

### Reproduction Link

https://codepen.io/anon/pen/RxKMxW
Selecting all when sorted by id works fine, but when sorting by name or anything else that breaks the order of the ID column, not all elements are selected, only around half. https://i.imgur.com/pSTfj6N.png

VdustR:
BREAKING CHANGE: no more `md-selected` event from `MdTableRow` because of too many duplicated events on a selection. It should be emitted only from `MdTable` both in `single` and `multiple` selecting mode. It was emitted from `MdTableRow` in `single` and from `MdTable` in `multiple` before.

# Feature

new props `:md-selected-value.sync` for reactive selection.

```vue
  <md-table v-model="people" md-card md-sort="name" :md-selected-value.sync="selected" @md-selected="onSelected">
```

Notice that disabled row could be selected from `:md-selected-value`.

fix #1292

# Refactor & fix

`MdTable.selectedItems`, `MdTable.selectable` change type to array for easier selection controlling.

Selection is determined by instance comparing now.

It was determined by comparing ids in single selection mode and comparing indexes in multiple selection mode before. The indexes way is not sorting the checkbox while sorting because the item is change and index is not. And it could also make duplicated items in `selectedItems` or `selectable`.

fix #1348

### DIFF
--- a/docs/app/pages/Components/Table/examples/TableSingle.vue
+++ b/docs/app/pages/Components/Table/examples/TableSingle.vue
@@ -1,11 +1,11 @@
 <template>
   <div>
-    <md-table v-model="people" md-card>
+    <md-table v-model="people" md-card @md-selected="onSelect">
       <md-table-toolbar>
         <h1 class="md-title">Selection Colors</h1>
       </md-table-toolbar>
 
-      <md-table-row slot="md-table-row" slot-scope="{ item }" :class="getClass(item)" md-selectable="single" @md-selected="onSelect">
+      <md-table-row slot="md-table-row" slot-scope="{ item }" :class="getClass(item)" md-selectable="single">
         <md-table-cell md-label="ID" md-sort-by="id" md-numeric>{{ item.id }}</md-table-cell>
         <md-table-cell md-label="Name" md-sort-by="name">{{ item.name }}</md-table-cell>
         <md-table-cell md-label="Email" md-sort-by="email">{{ item.email }}</md-table-cell>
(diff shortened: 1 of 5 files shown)


rayfoss:
Reactivity is a bit weird https://codesandbox.io/s/lp7117wylq it only works after next tick... should be built in

Issue #1357, opened 2017-12-29
[MdSelect] position


recallhyx:
### Steps to reproduce

the website [offical website](https://vuematerial.io/components/select/)
select the last of "font“ option ,after select again,the pop of option will go up 

### Which browser?
chrome

### What is expected?

the position of options should be fixed

### What is actually happening?
![image](https://user-images.githubusercontent.com/23056574/34433185-847346c4-ecb8-11e7-82a5-0b56174f92a3.png)

pop of options show the selected option position

AtiqDev:
I have taken following steps in order to set width and Position of a dropdown list content
**to position the dropdown right below the `<md-field>`**

> 1. give an id to the node **`<md-field>`** 

` <md-field id="field_apptStatus">`
        `<label class="at-focused">Appt Status</label>`
`             <md-select v-on:md-opened="apptStatusOpened" v-model="ptFilt.appt_status">`
`                     <md-option value="Scheduled">Scheduled</md-option>`
`                      <md-option value="No Show">No Show</md-option>`
`              </md-select>`
`     </md-field>`

> 2. Bind the emitted event **`md-opened`** to a method inside your vue component.
For instance my method that will be called upon **`md-opened`** is **`apptStatusOpened`**
obtain the _`...offset().top`_ and _`...offset().left`_ of _`<md-field>`_ and  Set those to _`top`_ and _`left`_ on _.md-select-menu_

`
            this.$nextTick(function () {`
`                this.$nextTick(function () {`
`                    this.$nextTick(function () {`
`                        $(".md-select-menu").css('top', ($("#field_apptStatus").offset().top + 60));`
`                        $(".md-select-menu").css('left', ($("#field_apptStatus").offset().left))`
`                    });`
`                });`
`            });`

> 3. In order to set width of the dropdown container equal to the width of the **`<md-field>`** we need to refactor the vue-material.css **`.md-menu-content{`** such as remove or comment out these two styles **`/*min-width: 112px;*/     /*max-width: 280px;*/`**. and then in the set the width of **_.md-select-menu_** upon **`md-opened`**  such as : `$j(".md-select-menu").css('width', ($j("#field_apptStatus")[0].offsetWidth))`

PR #1368, opened 2018-01-04
 refactor(MdSelect): remove uuid

VdustR:
Remove `uuid` from `MdMenuContent` and `MdSelect` and use `ref` instead of `uuid` to get menu contain container.

Field would be focused on the menu entered. I think it's better than the `setTimeout` way.


Samuell1:
Fix #1357

Issue #1278, opened 2017-12-07
[md-input] is not recognized in form-serialize


CharlesKumar:
vue.js latest stable (^2.5.8)
vue-material@beta latest (^1.0.0-beta-7)

[form-serialize](https://github.com/defunctzombie/form-serialize) is popular for serializing form inputs 

It process the input element correctly but returns nothing with md-input

> // node names which could be successful controls
var k_r_success_contrls = /^(?:input|select|textarea|keygen)/i;

Upon rendering md-input becomes input (inspected and copied from browser console)

`<div class="md-field md-layout-item md-size-48 m-1p md-small-size-100 md-theme-default md-has-value">
<label for="username">Your Name</label>
 <input id="username" class="md-input" type="text">   </div>`

How to approach this 

1. whether modify serialize function to see for md-input 

2. use vuejs hooks to determine post DOM render and call serialize

Expected behavior:

serialize the form data to use in ajax request (XMLHttp)

Samuell1:
I recommend using vuelidate to validate forms. You can see example here: https://vuematerial.io/components/form

CharlesKumar:
Thanks for quick reply 

The above examples doesn't contain information on how to serialize form data 

I am looking for serialize the form in client side and send them through ajax request to the server.

Based on response display success or failure message.

zapping:
Can you try by including the `name` attribute as well along with the `md-input` component? Like this:

```
<md-field>
    <label for="username">Your Name</label>
    <md-input id="username" name="username" v-model="initial"></md-input>
</md-field>
```

Samuell1:
@CharlesKumar vuelidate is cool package to validate or serialize you can see info about it here: https://monterail.github.io/vuelidate/

korylprince:
@CharlesKumar I've created an example of how you could use this library: https://codepen.io/korylprince/pen/LOwNro

However, you really shouldn't. You're much better off constructing the data object you want to send your self and using something like axios to serialize and do your XMLHttpRequest.

Using serialize makes you interact directly with the DOM which isn't really the Vue way, especially if there's a lot better alternatives.

CharlesKumar:
Thanks for the quick response and especially for an example in codepen 

your [example](https://codepen.io/korylprince/pen/LOwNro) has error both in firefox and chrome

> Uncaught ReferenceError: module is not defined

[...]

PR #1385, opened 2018-01-09
fix(MdFieldMixin) Preserve the name attribute on change

jastkand:
I've created a codepen which demonstrates an issue: https://codepen.io/jastkand/pen/EoQGKV

You can open a devtools and see that there are `name` attributes on `input` and `textarea`. When you try to change any input the `name` attribute will be removed which is not an expected behavior.

This also should fix this issue https://github.com/vuematerial/vue-material/issues/1278, because serialization uses `name` attributes

### DIFF
--- a/src/components/MdField/MdFieldMixin.js
+++ b/src/components/MdField/MdFieldMixin.js
@@ -2,6 +2,7 @@ export default {
   props: {
     value: {},
     placeholder: String,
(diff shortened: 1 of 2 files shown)


CharlesKumar:
when this change gonna be in the release (atleast in next beta)

PR #1394, opened 2018-01-11
fix(MdSteppers): overflow on resize

### LINKED ISSUE #1390: [MdSteppers] content overflows on resize
### Steps to reproduce

Putting a md-layout inside of a stepper. This layout has a couple of md-layout-items with the classes md-small-size-50 and md-xsmall-size-100. When the window is resized so that the content gets rearranged, the stepper doesn't resize, resulting in the overflowing content being hidden without even a scroll view to access it.
(The described layout is just an easy to reproduce example, but I've observed the same behavior on the most simple content, such as a couple of <p> tags, as is the case of the second or third step in the codepen at the bottom of this issue).

### Which browser?

Tested on Chrome.

### What is expected?

The stepper should resize according to the height of the content, resulting in the size the stepper is set to, once the page with the wrongly sized stepper is reloaded. Alternatively, the overflow property of the stepper should be set to auto, to keep the overflown content accessible.

The content overflows the stepper.

### Reproduction Link

Drag the separator to make the window smaller and see the overflown content, and reload to see the actually resized stepper.
https://codepen.io/MarinaGranell/pen/YYaGoL?editors=1010

Samuell1:
Fix #1390

### DIFF
--- a/src/components/MdSteppers/MdSteppers.vue
+++ b/src/components/MdSteppers/MdSteppers.vue
@@ -246,7 +246,12 @@
         this.setupObservers()
         this.setupWatchers()
       }, 100)
-    }
+    },
+    beforeDestroy () {
+      if (!('ResizeObserver' in window)) {
+        window.removeEventListener('resize', this.calculateStepperPos)
+      }
+    },
   })
 </script>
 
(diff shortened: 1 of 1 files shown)


VdustR:
This event listener should be removed `beforeDestroy` or `destroyed`

PR #1409, opened 2018-01-17
feat(MdDatepicker): custom first day of a week

### LINKED ISSUE #1397: [MdDatepicker] option for week to start on Monday instead of Sunday
Here in Europe, weeks start on Monday, so our clients would expect the datepicker to do that.

I could try to code this up myself and make a pull request, if you like?

VdustR:
new props `md-first-day-of-a-week` for customizing first day of a week.

fix #1397


marcosmoura:
I this shouldn't be a prop. Because for multiple Datepicker fields it will take a lot of effort to set the prop to it!
There is a global config that we can use.

What do you think? Can you change that?

Issue #1287, opened 2017-12-09
[MdDatepicker] font color with empty value in dark theme


VdustR:
### Steps to reproduce

Check the [official documentation](https://vuematerial.io/components/datepicker) and check any example with default value as empty, click "revert colors" button.

### Which browser?

Google chrome Version 62.0.3202.94 (Official Build) (64-bit), Xubuntu 17.04
`vue-material@1.0.0-beta-7`

### What is expected?

The text `dd-mm-yyyy` should be white.

### What is actually happening?

It's black.

![image](https://user-images.githubusercontent.com/29639463/33798111-1755a038-dd4e-11e7-86aa-3a89542a4f9f.png)

Works fine on firefox:
![image](https://user-images.githubusercontent.com/29639463/33814445-28dda5a4-de65-11e7-9869-91cd36fa068b.png)

### Reproduction Link

[official documentation](https://vuematerial.io/components/datepicker)

VdustR:
now we make the field empty while the value is not set to resolve the original problem with `vue-material@1.0.0-beta-7`

In this way, the empty value datepicker with label looks weird.

![image](https://user-images.githubusercontent.com/29639463/34360213-7cf917c8-ea99-11e7-9236-00562bdf2b50.png)

I think the label should have the same style like in input:

![image](https://user-images.githubusercontent.com/29639463/34360250-d1f81a26-ea99-11e7-8495-ef601cf9a65b.png)

VdustR:
Still the same while turn `mdOverrideNative` false

PR #1417, opened 2018-01-18
feat(MdDatepicker): remove fixed label position

Samuell1:
Resolves https://github.com/vuematerial/vue-material/issues/1287#issuecomment-353984365

### DIFF
--- a/src/components/MdDatepicker/MdDatepicker.vue
+++ b/src/components/MdDatepicker/MdDatepicker.vue
@@ -1,5 +1,5 @@
 <template>
-  <md-field class="md-datepicker">
+  <md-field :class="['md-datepicker', { 'md-native': !this.mdOverrideNative }]">
     <md-date-icon class="md-date-icon" @click.native="toggleDialog" />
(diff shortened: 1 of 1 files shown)


src/components/MdDatepicker/MdDatepicker.vue
```diff
@@ -127,9 +127,6 @@
   }
 
   .md-datepicker {
-    label {
-      top: 0 !important;
-    }
```
VdustR: We need a class to separate native and vue material datepicker dialog:

image

src/components/MdDatepicker/MdDatepicker.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <md-field class="md-datepicker">
+  <md-field :class="['md-datepicker', { 'md-native': !this.mdOverrideNative }]">
```
VdustR: I think this should be the opposite statement with `!isFirefox || this.mdOverrideNative`.

There could be a computed called `native`.

check here:
https://github.com/vuematerial/vue-material/blob/6a719b5c2421d2efadb3bc522693e438a5a01066/src/components/MdDatepicker/MdDatepicker.vue#L85

PR #1419, opened 2018-01-21
feat(MdRipple): multiple waves

Samuell1:

### DIFF
--- a/src/components/MdRipple/MdRipple.vue
+++ b/src/components/MdRipple/MdRipple.vue
@@ -7,9 +7,9 @@
     @mousedown.passive.stop="startRipple">
     <slot />
 
-    <transition name="md-ripple" @after-enter="clearWave" v-if="!isDisabled">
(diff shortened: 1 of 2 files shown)


src/components/MdRipple/MdRipple.vue
```diff
@@ -1,22 +1,35 @@
...
+    return function () {
+      clearTimeout(timeoutID)
+      var args = arguments
+      var that = this
+      timeoutID = setTimeout(function () {
+        fn.apply(that, args)
+      }, delay)
+    }
+  }
+
```
VdustR: I think this method could put it into [`/src/core/utils`](https://github.com/vuematerial/vue-material/tree/6a719b5c2421d2efadb3bc522693e438a5a01066/src/core/utils
) and name it `MdDebounce`.

It could be shorter with es6:

```javascript
  const debounce = (fn, delay) => {
    let timeoutID = null
    return (...args) => {
      clearTimeout(timeoutID)
      timeoutID = setTimeout(() => fn(...args), delay)
    }
  }
```

PR #1426, opened 2018-01-23
feat(MdDatepicker): emit date without confirming

### LINKED ISSUE #1424: [MdDatepicker] Throw events
I've had some people tell me that the way the datepicker saves the date on the bound variable is not intuitive, as you have to click on the "OK" button to actually select it. People have been clicking on the desired date and closing the datepicker by clicking outside of it, which doesn't actually save the selected date. Currently, there is no way to get the selected date but for the user to close the datepicker by clicking "OK".

Throwing an event onDateSelected with the clicked value and an event onDatepickerClosed would solve this problem, as these events could be listened to in order to reproduce the desired behavior, without actually changing how the datepicker works.

Samuell1:
Resolves #1424

### DIFF
--- a/src/components/MdDatepicker/MdDatepicker.vue
+++ b/src/components/MdDatepicker/MdDatepicker.vue
@@ -1,5 +1,5 @@
 <template>
-  <md-field class="md-datepicker">
+  <md-field class="md-datepicker" md-clearable>
     <md-date-icon class="md-date-icon" @click.native="toggleDialog" />
     <md-input :type="type" ref="input" v-model="modelDate" @focus.native="onFocus" pattern="[0-9]{4}-[0-9]{2}-[0-9]{2}" />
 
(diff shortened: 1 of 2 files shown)


VdustR:
Just an idea: would there be a **clear** button for clearing the value and keep **cancel** closing the dialog?

Samuell1:
I think cancel is okay, it should be like this because you *cancel* selecting a date not a *close* a datepicker

marcosmoura:
I think that should have a clear button instead. Something like the clear button with the 'md-clearable' on md-field. This will separate the responsibilities from closing the dialog without changing the date (maybe the user opened it by mistake or just to check if it's correct or not, by looking at the full calendar month), clearing the field form the proper clear button and confirming the new/old date. This will also be compliant with the guidelines!

What do you guys think? @Samuell1 @VdustR

VdustR:
Agree.

PR #1430, opened 2018-01-23
feat(MdDatepicker): add dateformat

### LINKED ISSUE #1422: [MdDatepicker] Dates are reversed in Safari
### Steps to reproduce
Go to https://vuematerial.io/components/datepicker
using Safari

### Which browser?
Safari on Osx Version 11.0.2 (12604.4.7.1.6)

### What is expected?
Date format should be dd/mm/yyyy

Date format is yyyy-mm-dd

Samuell1:
Resolves #1422

### DIFF
--- a/src/components/MdDatepicker/MdDatepicker.vue
+++ b/src/components/MdDatepicker/MdDatepicker.vue
@@ -96,9 +96,10 @@
       dateToHTMLString (date) {
         if (date) {
           let formattedDate = null
+          const dateFormat = this.$material.locale.dateFormat || 'YYYY-MM-DD'
 
           try {
-            formattedDate = format(date, 'YYYY-MM-DD')
+            formattedDate = format(date, dateFormat)
           } catch (error) {
             Vue.util.warn(`The datepicker value is not a valid date. Given value: ${date}.`, this)
           }


VdustR:
There are some problems with this:
* This pattern is not satisfied with other date formats like `'DD-MM-YYYY'`:
    https://github.com/vuematerial/vue-material/blob/70b3aa2bbf17f78f1bc00c4439d29db21a659ca8/src/components/MdDatepicker/MdDatepicker.vue#L4

* `parse` will occur error with `'DD-MM-YYYY'` format. It will got `Invalid Date` and occur infinite loop:
    https://github.com/vuematerial/vue-material/blob/70b3aa2bbf17f78f1bc00c4439d29db21a659ca8/src/components/MdDatepicker/MdDatepicker.vue#L75

* Even `<input type="date" />` shown as `DD/MM/YYYY` or else with some locale, but the value is still `YYYY-MM-DD`. It's better to control the date with one format and make it shown with custom format. For example: I always got `2018-01-24` from back-end API and only shown as `24/01/2018`, I still want to communicate with back-end with format `2018-01-24`. It's only show different strings for each end users depended on their locale.

PR #1448, opened 2018-01-26
fix(MdTabs): fix indicator

### LINKED ISSUE #1304: [MdTabs] indicator has wrong behaviour in Safari
### Steps to reproduce

Go here: https://vuematerial.io/components/tabs/home with Safari on OSX and click the tabs

### Which browser?

Safari 11.0.2 (12604.4.7.1.4) OSX Sierra 10.12.6 (16G1036)

### What is expected?

The tab indicator should slide gracefully back and forth when I click the tabs.

The indicator slides back and forth, but the width varies from 0 to much-too-wide

This seems to be an issue with transitioning the value for "right" on the tab indicator. The issue is not appearing in Chrome or Firefox, just Safari.

VdustR:
``right: `calc(100% - ${buttonWidth + buttonLeft}px)` `` work weird with safari, replace it with `width` way to set indicator position

fix #1304

### DIFF
--- a/src/components/MdTabs/MdTabs.vue
+++ b/src/components/MdTabs/MdTabs.vue
@@ -191,7 +191,7 @@
 
             this.indicatorStyles = {
               left: `${buttonLeft}px`,
-              right: `calc(100% - ${buttonWidth + buttonLeft}px)`
+              width: `${buttonWidth}px`
             }
           }
         })


marcosmoura:
We can use this for now, but the transition of the indicator do not follow the guidelines with this change.

PR #1457, opened 2018-01-29
fix(MdLayout): remove breakpoint margins

### LINKED ISSUE #1307: md-toolbar changes size at incorrect widths
### Steps to reproduce

open https://vuematerial.io/components/app  
Resize the screen and note exactly where the toolbar changes height.

### Which browser?

At least chrome

### What is expected?

The toolbar should changes size at 600 and 960 (as per https://material.io/guidelines/layout/responsive-ui.html#responsive-ui-breakpoints )

the toolbar is resizing at 600 and 944 due to https://github.com/vuematerial/vue-material/blob/652a0c684d336227cb55a8f5f159bdb6161d214c/src/components/MdLayout/mixins.scss#L15

This line of code appears to be compensating for margins or something, but it doesn't seem to be necessary in this case.

### Reproduction Link

https://vuematerial.io/components/app

Samuell1:
Fix #1307

### DIFF
--- a/src/components/MdLayout/mixins.scss
+++ b/src/components/MdLayout/mixins.scss
@@ -12,13 +12,13 @@
 }
 
 @mixin md-layout-small {
-  @media (max-width: #{$md-breakpoint-small - 16px}) {
+  @media (max-width: #{$md-breakpoint-small}) {
     @content;
   }
 }
 
 @mixin md-layout-medium {
-  @media (max-width: #{$md-breakpoint-medium - 16px}) {
+  @media (max-width: #{$md-breakpoint-medium}) {
     @content;
   }
 }


zevdg:
Shouldn't https://github.com/vuematerial/vue-material/blob/612a63d4f7e9908e2ed0a4a2986bd7a542357ce9/src/components/MdLayout/mixins.scss#L27 and https://github.com/vuematerial/vue-material/blob/612a63d4f7e9908e2ed0a4a2986bd7a542357ce9/src/components/MdLayout/mixins.scss#L33 also have been fixed in this PR?  They appear to have the same problem.

marcosmoura:
@zevdg Large breakpoints should have this 16px margin, as stated in the guidelines:

> Desktop breakpoints are 16dp below the listed values to accommodate variations in browser.

zevdg:
Aaaahh, I see it now.  Thanks for the clarification.

PR #1460, opened 2018-01-31
fix: revert old babel

VdustR:
avoid error:

```text
This dependency was not found:

* @babel/polyfill in multi @babel/polyfill ./docs/app/index.js ./build/local/client
```


VdustR:
fixed a8e3978fca34311cdb8ceb815d1b49660883b993

PR #1462, opened 2018-01-31
refactor(MdMenu): rework keynavigation

Samuell1:
Refactor keyboard navigation to global events because elements cant listen to keyboard.


marcosmoura:
@Samuell1 Can you merge your branch with dev?

src/components/MdMenu/MdMenuContent.vue
```diff
@@ -206,10 +203,30 @@
...
         }
       },
+      createKeydownListener () {
+        window.addEventListener('keydown', this.keyNavigation)
+      },
+      destroyKeyDownListener () {
+        window.removeEventListener('keydown', this.keyNavigation)
+      },
+      keyNavigation (event) {
+        switch (event.key) {
```
marcosmoura: If the page has scrollbars, the event is not working as it should be. Maybe adding a `event.preventDefault()` can prevent window scrollbars from scroll.

PR #1496, opened 2018-02-08
fix(MdTableRow): update selectableItem while `mdItem` changed

### LINKED ISSUE #1492: md-table's data doesn't get updated properly

I'm trying to implement a pagination function with vue 2.5.3 / vue-material1.0.0 and the problem is whenever I update the
model , the table gets updated with the newly fetched model but the md-table component still has the previous data so when I check the "select all" box it gets the previous data only which makes the selected variable contain only the page 1 data when in fact I am in page 2. anyone have any suggestions on how to deal with this?

VdustR:
fix #1492

### DIFF
--- a/src/components/MdTable/MdTableRow.vue
+++ b/src/components/MdTable/MdTableRow.vue
@@ -83,6 +83,10 @@
       },
       mdSelectable () {
         this.MdTable.selectingMode = this.mdSelectable
+      },
+      mdItem (after, before) {
+        this.removeSelectableItem(before)
+        this.$nextTick(this.addSelectableItem)
       }
     },
     methods: {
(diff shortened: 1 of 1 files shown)


ueharajohji:
Hi , guys sorry for the late updated. Will 'npm update' get this fix since it is already committed ?

Samuell1:
No, it needs to be released @ueharajohji

ueharajohji:
In order to use the fix ,I just need to change the file manually and rebuild , is that correct?

Samuell1:
Yes @ueharajohji

ueharajohji:
I'm sorry for the lack of research but I tried running the command 'npm install' to get the node modules and  'npm run -script build' to get a new build file but it shows me an error when building.

ueharajohji:
Hello , I successfully build the file , thanks again for the fix.

PR #1498, opened 2018-02-08
Documentation: Created controlled radio example. Meta: Added dist dir…

brianboyko:
Added a controlled radio example to documentation, showing primarily how one would use Vuex actions/getters to control a set of radio buttons.  (Even though the documentation does not use vuex, it uses a method to create the same effect.) 

Also added the dist/ directory to gitignore (seems like an oversight) and added a few more clarifying examples to the primary radio button example. 
screen shot 2018-02-08 at 11 01 05 am

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage/
 yarn-debug.log
 npm-debug.log
 .vscode
+dist/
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@ coverage/
 yarn-debug.log
 npm-debug.log
 .vscode
+dist/
```
VdustR: don't ignore this or unpkg link will be dead

marcosmoura:
@brianboyko The documentation uses Vuex. You can use that on your example.
As @VdustR said, do not ignore dist folder.

marcosmoura:
@brianboyko This is a good example, but we also have to add the same example for Checkbox and Switch. Can you do that?

Samuell1:
@brianboyko Any updates to this PR?

Issue #1500, opened 2018-02-09
[MdTabs] content length calculations


mrwhy-orig:
### Steps to reproduce
Use md-tabs with long content, in my case several md-cards in a tab within a sub tab.
I'm using vue-material beta 7 (edit) and 8

### Which browser?

FF (v. 58) and Chrome (v. 64)

### What is expected?
Displaying the whole content. Scrollable in the full height.

### What is actually happening?
The outer tab height is not equal to the inner tab height. Actually in my setup it is much smaller (inner tab height = 2489px, outer = 1748px)

VdustR:
Would you provide your code in [Codepen](https://codepen.io/)?

mrwhy-orig:
https://codepen.io/MrWhy/pen/YeVRZw

emman27:
Hey! I'm facing similar issues.

VdustR:
This reproduction act differently between `1.0.0-beta-7` and `1.0.0-beta-8`. I got CPU leakage with your code in `1.0.0-beta-8` and I fixed that with #1506. Would you kindly to check whether if this is fixed with #1056?

mrwhy-orig:
I tried your fix without any luck. With your code, there is another issue. If i use the route url directly the sub tabs are not getting rendered anymore. If I use the index url and navigate to the page the sub tabs are getting rendered.

PR #1506, opened 2018-02-13
fix(MdRipple): clear immediately

VdustR:
* clear wave immediately instead of debounce delay.

* `<transition group>` make `MutationObserver` callback many times by one click, and it will cause page dead with CPU leakage.
    This could reproduce with this reproduction: https://github.com/vuematerial/vue-material/issues/1500#issuecomment-364850236 with `1.0.0-beta-8`.
    I can't find a way to configure the `<transition group>` to work as fine as `<transition>`. The solution for now is using `<transition>` and make waves act like old single ripple with `animating`.

related to #1419, #1500


VdustR:
fix #1700

PR #1525, opened 2018-02-18
fix(MdSelect): re-fix select content under dialog

### LINKED ISSUE #1149: [md-select] Select is behind dialog
Version: Beta 1.0.4/1.0.3

bildschirmfoto vom 2017-11-16 20-01-40

This seems to fix it:

```
.md-menu-content {
  z-index: 10;
}
```

onjiro:
Issue #1149 has been reproducted on v1.0.0-beta-8, after commit f0a7325 .
CSS specificity should not same as MdMenuContent.
It seems z-index on MdMenuContent overriding.
https://github.com/vuematerial/vue-material/blob/61296b366c80e47b3a36b0acd6e4e84045e0b4da/src/components/MdMenu/MdMenuContent.vue#L253-L265 

fix #1149 
thanks.

### DIFF
--- a/src/components/MdField/MdSelect/MdSelect.vue
+++ b/src/components/MdField/MdSelect/MdSelect.vue
@@ -329,18 +329,16 @@
       border: 0;
     }
   }
-  .md-menu-content {
+  .md-menu-content.md-select-menu {
     z-index: 111;
-    &.md-select-menu {
-      width: 100%;
+    width: 100%;
 
-      &.md-menu-content-enter {
-        transform: translate3d(0, -8px, 0) scaleY(.3);
-      }
+    &.md-menu-content-enter {
+      transform: translate3d(0, -8px, 0) scaleY(.3);
+    }
 
-      .md-list {
-        transition: opacity .3s $md-transition-drop-timing;
-      }
+    .md-list {
+      transition: opacity .3s $md-transition-drop-timing;
     }
   }
 </style>


Samuell1:
This is not a fix because you added same class to `.md-menu-content` that is few lines under it as `&.md-select-menu`

onjiro:
Sorry, I missed it. I'll fix.

Samuell1:
And i tested now select in dialog in dev branch and all things works.

onjiro:
Thank you for testing and notice.
Hmmm...but, this problem seems still existing in dev branch.
I tried `npm run build` and tested on dev branch. 
If you tested on other way, could you try this?

Samuell1:
When you try with `npm run dev` without fix it works?

onjiro:
OK. It works fine with `npm run dev` without fix.

This issue is caused by same CSS specificity.
It makes components flagile. So I think to fix CSS is better than to fix build process. thanks.

Samuell1:
@onjiro CSS is fixed, but build process is removing, changing some CSS and thats why it doesnt work. @marcosmoura Can you look at it?

marcosmoura:
@onjiro @Samuell1 I will take a look at it today.

JRakhimov:
So is there any solutions for this problem?

marcosmoura:
@JRakhimov Not yet. We will upgrade the build system, so this problem will be vanished.

PR #1540, opened 2018-02-23
Add event handling to input field

jkosonen:
Add event handling to input field


VdustR:
We don't need this because it's already bound listeners

https://github.com/vuematerial/vue-material/blob/156506ba83080c75409d517f17ff21cc370338fa/src/components/MdField/MdInput/MdInput.vue#L6

Issue #1206, opened 2017-11-23
[MdSelect] wrong position if body has margin


hurradieweltgehtunter:
### Steps to reproduce

set 
```
body { margin-top: 50px }
```

### Which browser?

Chrome, Firefox, but not relevant

### What is expected?

Clicking on a md-select should open the options right below or on the select field (like in docs).

### What is actually happening?

clicking on a md-select opens the options on a wrong position
![image](https://user-images.githubusercontent.com/16665512/33165685-055604fe-d038-11e7-8ea5-046b52ae7b1e.png)

Removing the margin on body opens the options on correct position.
![image](https://user-images.githubusercontent.com/16665512/33165809-71d45e3c-d038-11e7-8cb2-71cc6708884c.png)

marcosmoura:
This is a bad bug. I will fix asap

hurradieweltgehtunter:
fyi: I replaced margin by padding and it works fine. nevertheless, bug :)

PR #1551, opened 2018-02-24
[MdSelect] Adjust position when body has margin

AlbertLucianto:
This PR is a possible fix for issue #1206.

## Description
This issue might be related to Popper.js library.

After render, `popper` instance which wraps `MdMenuContent` is mounted as a direct child of `body`, and changes in `body`'s position affects where the popper is mounted.

By adjusting `popperSettings.modifiers.offset` with `body`'s position relative to the viewport, the menu will be correctly positioned.

### Style
```css
body {
  margin-top: 100px;
  margin-left: 100px;
}

/* Or an element which can affect body's position */
#app {
  margin-top: 100px;
  margin-left: 100px;
}
```
### Before
Menu is off position the size of the `body`'s `margin` (or any elements pushing the `body`s position).

screen shot 2018-02-24 at 7 18 25 pm

### After

screen shot 2018-02-24 at 7 16 38 pm

Really appreciate for feedbacks and reviews.

### DIFF
--- a/docs/app/pages/Components/Menu/Menu.vue
+++ b/docs/app/pages/Components/Menu/Menu.vue
@@ -175,20 +175,20 @@
         ]
       },
       events: {
-          headings: ['Name', 'Description', 'Value'],
-          props: [
-            {
-              name: 'md-opened',
-              description: 'Triggered when menu opens',
-              value: 'null'
-            },
-            {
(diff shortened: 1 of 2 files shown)


marcosmoura:
Sounds good for me! Thank you so much!

PR #1552, opened 2018-02-25
fix(MdTabs): remove ResizeObserver

### LINKED ISSUE #1500: [MdTabs] content length calculations

### LINKED ISSUE #1513: [MdTabs] content height is not set with async data
### Steps to reproduce

When filling a MdList into MdTab content from an async source, the height is not set correctly for initial tab.
eg I have this in my template.

`events` is filled in vuex, after calling an axios.get(...).

```
<md-tabs class="md-transparent" md-alignment="fixed">
                <md-tab id="tab-list-view" md-label="lijst" md-icon="list">
                    <div class="md-content">
                        <md-list>
                            <md-list-item :to="'/hikes/' + event.id" class="md-double-line" v-for="event in events"
                                          :key="event.id">
                                <div class="md-list-item-text">
                                    <ul class="meta md-layout">
                                        <li>
                                            <md-icon>today</md-icon>
                                            {{event.takes_place_on | shortDate}}
                                        </li>
                                        <li>
                                            <md-icon>location_on</md-icon>
                                            op {{event.start_location | fromPosition(currentLocation) | convert('m',
                                            'km', 1,
                                            true)}}
                                        </li>
                                        <li>
                                            <md-icon>directions_walk</md-icon>
                                            {{event.distances[0] | convert('m', 'km', 2, true)}}
                                        </li>
                                    </ul>
                                    <div class="description text-primary">
                                        {{event.name}} ({{event.organised_by_club.name}})
                                    </div>
                                </div>
                            </md-list-item>
                        </md-list>
                    </div>
                </md-tab>
                <md-tab id="tab-calendar-view" md-label="kalender" md-icon="today">
                    <div class="md-content"></div>
                </md-tab>
</md-tabs>

```

Initial view is this: (where md-tabs-content is set to style="height: 16px;")

When changing tab and return back to this tab:

 

### Which browser?

Vue v2.5.13
Vue-material 1.0.0-beta-8

Any browser

### What is expected?

When content is changed, a redraw should happen, so correct height is generated.

Content height was computed before async data was loaded. 
So height is set to 16px;

Samuell1:
Fix #1513 

Maybe it can resolve #1500 ?

### DIFF
--- a/src/components/MdTabs/MdTabs.vue
+++ b/src/components/MdTabs/MdTabs.vue
@@ -207,22 +207,20 @@
           }
         }
       },
+      callResizeFunctions () {
+        this.setIndicatorStyles()
+        this.calculateTabPos()
+      },
       setupObservers () {
-        if ('ResizeObserver' in window) {
-          this.resizeObserver = new window.ResizeObserver(this.setIndicatorStyles)
-          this.resizeObserver.observe(this.$el)
-        } else {
-          this.resizeObserver = MdObserveElement(this.$el.querySelector('.md-tabs-content'), {
-            childList: true,
-            characterData: true,
-            subtree: true
-          }, () => {
-            this.setIndicatorStyles()
-            this.calculateTabPos()
-          })
+        this.resizeObserver = MdObserveElement(this.$el.querySelector('.md-tabs-content'), {
+          childList: true,
+          characterData: true,
+          subtree: true
+        }, () => {
+          this.callResizeFunctions()
+        })
 
-          window.addEventListener('resize', this.setIndicatorStyles)
-        }
+        window.addEventListener('resize', this.callResizeFunctions)
       },
       setupWatchers () {
         if (this.mdSyncRoute) {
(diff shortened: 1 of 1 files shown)


mrwhy-orig:
I can confirm that this fix + the fix #1506 is solving a part of #1500. But there is one case left. If there is a child md-tabs with only 1 md-tab it is not calculating the height.
I tried to demonstrate it with a pen, you have to use the "debug" view because the whole fix isn't working in editor view.
[https://codepen.io/MrWhy/pen/QQZZme?editors=1111](https://codepen.io/MrWhy/pen/QQZZme?editors=1111)

marcosmoura:
@Samuell1 @mrwhy-orig We can merge this one and work on another fix for this case. Will be better.

mrwhy-orig:
@marcosmoura should I open another issue on this?

PR #1564, opened 2018-02-28
fix(MdRadio): make required actually work

williamchong:
Fixes the MdRadio not working in `form` when using `required` attr

Before:
no UI response or whatsoever on a form submit fail due to `required` `md-radio` not selected.
image

After
proper browser hint
image

Also since any `radio` with `required` affected all `radio` in with the same `name`, the extra * in css is actually confusing, so I removed it

Please comment if any change is needed

### DIFF
--- a/src/components/MdRadio/MdRadio.vue
+++ b/src/components/MdRadio/MdRadio.vue
@@ -2,7 +2,7 @@
   <div class="md-radio" :class="[$mdActiveTheme, radioClasses]">
     <div class="md-radio-container" @click.stop="toggleCheck">
       <md-ripple md-centered :md-active.sync="rippleActive" :md-disabled="disabled">
-        <input type="radio" v-bind="{ id, name, disabled, required, value }">
+        <input type="radio" :checked="isSelected" v-bind="{ id, name, disabled, required, value }">
       </md-ripple>
     </div>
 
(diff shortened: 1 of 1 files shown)


marcosmoura:
I don't think that the default message of HTML5 validations is good in this case. That's because it may vary in different browsers and will not work in others. I think, the best way, would be a javascript validation, showing a `md-error` class. I know that this may sound a little strange, but in our ideas, this should not be part of the framework itself.

williamchong007:
Since radio with same `name` can span across different locations in a `form`, we probably need a `md-form` or `md-radio-group` to perform javascript based validation ?

Browser validation message indeed can be tricky, but currently other `md-input` validation is depending on it too. Maybe you want this as a temp solution and to synchronize radio behavior with other input elements, then figure out how you want to do validation all together later on?

https://github.com/vuematerial/vue-material/pull/1564/files#diff-30887f9b93a1eda43705a6d32dc02f3aR5
Probably still want to merge this line if you really don't like the css part, it is effectively blocking any `form` with a `required` `md-radio` from able to submit at all.

marcosmoura:
I think, for now, a good approach to do a validation will be `novalidate` parameter on form tag. And then the dev will do a javascript validation integrating with the md-error class. I will keep this PR opened until we figure it out how to apply this in a more elegant way.

PR #1566, opened 2018-03-01
Emit event when clicking outside of dialog & add object to possible value to radio

jkosonen:
Emit event when clicking outside of dialog & add object to possible value to radio


Samuell1:
Why is with this change other files changed too?

Samuell1:
And i forget to mention for next PRs, commits please use this naming

https://github.com/vuematerial/vue-material/blob/dev/.github/COMMIT_CONVENTION.md

PR #1567, opened 2018-03-01
comp(MdDialog): Add emit event when clicked outside of dialog

jkosonen:
There might be cases where user wants to know if clicked outside of dialog (at least I have use cases)

### DIFF
--- a/docs/app/pages/Components/Dialog/Dialog.vue
+++ b/docs/app/pages/Components/Dialog/Dialog.vue
@@ -125,7 +125,7 @@
               },
               {
                 name: 'md-clicked-outside',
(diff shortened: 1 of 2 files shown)


marcosmoura:
@jkosonen I thing it's a good idea, since it do not interfere in any feature. Can you add this inside the documentation page of Dialog?

jkosonen:
@marcosmoura like that?

jkosonen:
And I think it's useful. I'm using it currently to ask confirmation from user

marcosmoura:
@jkosonen Sounds great. Thank you!

PR #1576, opened 2018-03-03
fix(MdListItemDefault): ripple effect if mdRipple prop is set to true or if the item has controls

lucasfernog:


marcosmoura:
@lucasfernog This PR will break the case when we want to have an event on the list-item but still have a control inside, for separated actions, like open wifi settings when clicking on the list item itself and toggle the wifi state. In that case it'll always toggle the control regardless.

PR #1577, opened 2018-03-04
feat(MdDatepicker): add close, open events

### LINKED ISSUE #1574: [MdDatepicker] Add close event
Our use case is to perform some action using the selected date upon closing of the datepicker dialog.  Currently we must use a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) on the html body to detect if the dialog div has been added or removed.  Adding $emit of an md-closed event from the MdDatepicker.vue when the showDialog flag is set to false would remove the need for our workaround.  It would also be great if we could differentiate between a close triggered by the OK button versus a close triggered by the CANCEL button or a click off of the dialog.

Samuell1:
Resolves #1574

### DIFF
--- a/src/components/MdDatepicker/MdDatepicker.vue
+++ b/src/components/MdDatepicker/MdDatepicker.vue
@@ -84,6 +84,11 @@
       toggleDialog () {
         if (!isFirefox || this.mdOverrideNative) {
           this.showDialog = !this.showDialog
+          if (this.showDialog) {
+            this.$emit('md-opened')
+          } else {
+            this.$emit('md-closed')
+          }
         } else {
           this.$refs.input.$el.click()
         }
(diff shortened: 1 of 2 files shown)


src/components/MdDatepicker/MdDatepicker.vue
```diff
@@ -84,6 +84,11 @@
       toggleDialog () {
         if (!isFirefox || this.mdOverrideNative) {
           this.showDialog = !this.showDialog
+          if (this.showDialog) {
+            this.$emit('md-opened')
+          } else {
+            this.$emit('md-closed')
+          }
```
VdustR: I think `md-opened` and `md-closed` for transition `enter` and `leave` is better.

How do you think?

marcosmoura: I think it's not needed. In this case, the md-opened and md-closed will fire an intention of close and open.

PR #1584, opened 2018-03-06
feat(MdTableCell): add `md-id` prop

### LINKED ISSUE #1535: [MdTable] autogenerated headers are not differentiated in any way
When using custom template table as we cannot sort otherwise headers are automatically generated with use md-label and md-sort-by.
Can we somehow add ids for them so that they can be differentiated from one another? Because frankly it causes quite a lot of problems for UI automated testing.
So it would be great if you guys either added support for md-sort-by for md-table-head or added generating of ids for custom template headers, thanks.

Samuell1:
Resolves #1535

### DIFF
--- a/src/components/MdTable/MdTableCell.vue
+++ b/src/components/MdTable/MdTableCell.vue
@@ -10,6 +10,7 @@
   export default {
     name: 'MdTableCell',
     props: {
+      mdId: [String, Number],
       mdLabel: String,
       mdNumeric: Boolean,
       mdTooltip: String,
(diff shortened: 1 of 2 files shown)


marcosmoura:
@Samuell1 Can you add a mention on the documentation website?

Samuell1:
@marcosmoura yes but there are missing docs fully :D

marcosmoura:
@Samuell1 What?

Samuell1:
@marcosmoura i mean Api is "Coming soon..."

PR #1585, opened 2018-03-06
fix(MdTable): emit selected / update event only when selected item(s) really changed

### LINKED ISSUE #1559: [MdTable] emitting @md-selected with no value
### Steps to reproduce
Using vue-material@1.0.0-beta-8, with the following code.
```
<template>
  <div>
    <md-table md-card v-model="lists" v-if="lists.length !== 0"  @md-selected="onSelect">
      <md-table-row slot="md-table-row" slot-scope="{ item }" md-selectable="single">
        <md-table-cell md-label="Name">{{ item.name }}</md-table-cell>
      </md-table-row>
    </md-table>
  </div>
</template>

<script>
import getLists from '@/api';

export default {
  data() {
    return {
      lists: [],
    };
  },
  async created() {
    try {
      this.lists = await getLists();
    } catch (err) {
      console.log(err)
    }
  },
  methods: {
    onSelect(item) {
      console.log(item);
    },
  },
};
</script>
```

### Which browser?
Multiple (Chrome 64, Firefox)

### What is expected & What is actually happening?
onSelect should only be called when something is clicked. However, it is being called every time the component appears (not sure created or mounted, tbh), and called with undefined.

VdustR:
fix #1559

### DIFF
--- a/src/components/MdTable/MdTable.vue
+++ b/src/components/MdTable/MdTable.vue
@@ -229,21 +229,11 @@
 
           if (isValEmpty && isOldEmpty) {
             return false
-          }
-
-          if (!isValEmpty && !isOldEmpty) {
-            if (val.length !== old.length) {
-              return true
-            }
-
-            if (val.every((item, index) => item == old[index])) {
-              return false
-            }
-
+          } else if (!isValEmpty && !isOldEmpty) {
+            return (val.length !== old.length) ? true : !val.every((item, index) => item == old[index])
+          } else {
             return true
           }
-
-          return true
         })()
 
         if (changed) {
--- a/src/components/MdTable/MdTable.vue
+++ b/src/components/MdTable/MdTable.vue
@@ -222,11 +222,38 @@
           this.MdTable.hasValue = this.hasValue
         }
       },
-      'MdTable.selectedItems' (val) {
-        this.select(val)
+      'MdTable.selectedItems' (val, old) {
+        let changed = (() => {
+          let isValEmpty = !val || val.length === 0
+          let isOldEmpty = !old || old.length === 0
+
+          if (isValEmpty && isOldEmpty) {
+            return false
+          }
+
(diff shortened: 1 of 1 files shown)


VdustR:
Additionally, I want to make another PR to implement model comparing in [utils](https://github.com/vuematerial/vue-material/tree/9d04824e897b16de789c3d579c7c0ca88b028bda/src/core/utils). Maybe named `MdModelComparator`?

All components have selection could call the function to check the value or array is changed or not, including `MdSelect`, `MdTable`, `MdRadio`.

for single comparing, I want to implement a method `compare(a, b)` followed these rules:

* `undefined` = `null`
* `null` != `0` != `NaN` != `''`
* `0` = `'0'`, `1` = `'1'`, ...
* any non-object-value != any object, an object != another object

for multiple comparing, I want to implement a method `compareArray(a, b)` followed these rules:

* `[]` = `null` = `undefined`
* regard sorting,`['a', 'b']` = `['b', 'a']`, including object, I'll using find index to implement that.
* values in array follow single comparing rule, like: `[null]` = `[undefined]`

These rules could be write in documentation to let developers know how it work.

@marcosmoura , @Samuell1
How do you think?

src/components/MdTable/MdTable.vue
```diff
@@ -222,11 +222,28 @@
           this.MdTable.hasValue = this.hasValue
         }
       },
-      'MdTable.selectedItems' (val) {
-        this.select(val)
+      'MdTable.selectedItems' (val, old) {
```
marcosmoura: I think this method is way complex. Maybe we can break it in some small functions to better describe it.

marcosmoura: Like: `val.length !== old.length) ? true : !val.every((item, index) => item == old[index]`

marcosmoura:
@VdustR I think it would be good. I want to make less code repetition.

PR #1591, opened 2018-03-10
remove mapCoverage jest config property from package.json (deprecated)

dennythecoder:
Removes deprecated warning when running tests.


dennythecoder:
Travis seems to fail on linting of unrelated component.  Something I need to redo?

PR #1600, opened 2018-03-13
fix: icons in selects are aligned to top

shafimsp:
This PR will fix icons in selects are aligned to top


marcosmoura:
@shafimsp Sounds great. Thank you so much!

Samuell1:
This will fix a selects issue but adds other issues with textarea and inline input

obrazok

PR #1601, opened 2018-03-13
fix(MdField): fix padding issues

Samuell1:
Fixed issues caused by #1600

### DIFF
--- a/src/components/MdField/MdField.vue
+++ b/src/components/MdField/MdField.vue
@@ -397,7 +397,12 @@
       }
     }
 
+    &.md-autogrow {
+      padding-top: 0;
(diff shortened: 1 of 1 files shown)


marcosmoura:
@Samuell1 The autogrow textarea is still misaligned

shafimsp:
I will take of it.. 
And also have some corrections on other field components

Ok?

marcosmoura:
@shafimsp I can't. It should be fixed now.
The other branches will need to be fixed as well.

shafimsp:
@marcosmoura ok, so may i merge this branch with branch which i work on?

Issue #1423, opened 2018-01-22
[MdTabs] pagination/scrollable tabs


hanxue:
### Steps to reproduce

1. Create many tabs so that all the tabs cannot fit into the width of a single line
2. The tab title will start to overwrite each other

![image](https://user-images.githubusercontent.com/760008/35233791-122abbda-ffda-11e7-81f1-ec9aeb4ca59c.png)

### Which browser?

Chrome: 65.0.3323.0
Firefox: 59.0a1

### What is expected?

The tabs should go to the next row. Example: (Bootstrap) http://jsbin.com/xesugox/edit?html,output

![image](https://user-images.githubusercontent.com/760008/35233816-2456c27c-ffda-11e7-96b6-12b838ac4ae8.png)

Samuell1:
Tabs now are missing scrolling if label is longer, thats why you cant use it for longer labels.

Material guidelines: https://material.io/guidelines/components/tabs.html#tabs-usage

igor-ribeiro:
It used to work: #267

Samuell1:
Yes, in old version works. @igor-ribeiro

PR #1618, opened 2018-03-21
feat: add swipeable functionality

### LINKED ISSUE #1539: [MdDrawer] md-swipeable
Hi,
Is there a plan to add **md-swipeable** of "md-sidenav" from the old 0.8 version to the new "md-drawer" in 1.0 version ?
Thank you.

**Vue Material is the best**

Samuell1:
Resolves #1539

ezgif-5-4b2a27581d

#### Components that support swipeable:
- [x] Drawer
- [x] Tabs

#### Usage of mixin in custom components:
Importing mixing to component.

```js
import MdSwipeable from 'core/mixins/MdSwipeable/MdSwipeable'

// passing it as a mixin to component
mixins: [MdSwipeable],
```

```js
watch: {
      swiped (value) {
         // value of swiped direction (left, right, top, bottom)
      }
}
```

```js
computed: {
      mdSwipeElement () {
         return this.$el // element on what will be swipe detected, if not provided default is used as window
      }
}
```

Props:
`md-swipeable` - enables swipeable events

```js
mdSwipeThreshold: {
    type: Number,
    default: 50
},
mdSwipeRestraint: {
    type: Number,
    default: 100
},
mdSwipeTime: {
    type: Number,
    default: 400
}
```


zapping:
@Samuell1 does this fix #1423 also.

Samuell1:
@zapping No, that needs to be implemented separately, but we can use same mixin for touch support

src/components/MdDrawer/MdDrawer.vue
```diff
@@ -11,8 +11,11 @@
   import MdOverlay from 'components/MdOverlay/MdOverlay'
   import MdPropValidator from 'core/utils/MdPropValidator'
 
+  import MdSwipeable from 'core/mixins/MdSwipeable/MdSwipeable'
+
   export default new MdComponent({
     name: 'MdDrawer',
+    mixins: [MdSwipeable],
```
marcosmoura: For the drawer element, we should use the swipeElement on drawers. It can't be window, because we need to ensure those right drawers will work.

src/core/mixins/MdSwipeable/MdSwipeable.js
```diff
@@ -0,0 +1,87 @@
...
+    swipeStartTime: null,
+    swiped: null,
+    touchPosition: {
+      startX: 0,
+      startY: 0
+    }
+  }),
+  computed: {
+    getSwipeElement () {
+      return this.swipeElement || window
```
marcosmoura: I think the `swipeElement` should have the `md` prefix

marcosmoura:
@Samuell1 Can you add the props description to docs page?

mismith:
Anything I can do to help?

sinisarudan:
Hi, is this still not solved?
Namely, even after adding the prop:
<md-tabs md-swipeable>, md-tabs looks like this.
tabs being cut-out, squeezed and not swipeable.
image

Issue #1182, opened 2017-11-20
[core] regeneratorRuntime is not defined


jackovsky888:
I poorly just don't get this to run.

### Steps to reproduce

As soon as I import -
`import { MdApp } from 'vue-material/dist/components'`
`import 'vue-material/dist/vue-material.min.css'`
`Vue.use(MdApp)`

to my project, I get:

`index.js:1 Uncaught ReferenceError: regeneratorRuntime is not defined
    at index.js:1
    at Object../node_modules/vue-material/dist/components/index.js.Object.defineProperty.value (index.js:1)`

I am working with JeffreyWay/laravel-mix
`let mix = require('laravel-mix');
mix.js('resources/assets/spa/main.js','public/js/app')`

Happens also if i try to import the whole bundle or other components.

flyinghawker:
same issue here, just try to set up vue-material in my plain new project but this annoy me, no way to figure out.

Samuell1:
Looks like its because you doesnt have await/async: https://stackoverflow.com/a/33527883

jackovsky888:
@Samuell1 - Thanks! This actually was the solution...

That is now my package.json:
`    "devDependencies": {`
`        "axios": "^0.17",`
`        "babel-core": "^6.0.20",`
`        "babel-plugin-transform-runtime": "^6.23.0",`
`        "babel-polyfill": "^6.0.16",`
`        "babel-preset-env": "^1.6.1",`
`        "babel-preset-stage-3": "^6.24.1",`
`        "bootstrap-sass": "^3.3.7",`
`        "cross-env": "^5.1",`
`        "jquery": "^3.2",`
`        "laravel-mix": "^1.0",`
`        "lodash": "^4.17.4",`
`        "vue": "^2.1.10"`
`    },`
`    "dependencies": {`
`        "babel-runtime": "^6.26.0"`
`    }`

my .babelrc:
`  "presets": [ "env", "stage-3" ],`
`  "plugins": [`
`    ["transform-runtime", {`
`      "polyfill": false,`
`      "regenerator": true`
`    }]`
`  ]`

and on the top of my index.js i added:
`import "babel-core/register"`
`import 'babel-polyfill'`

marcosmoura:
This is quite strange. I will take a deep look. The babel stuff should be built in Vue Material.

marsvin:
Maybe the transform-async-to-generator transform could be an alternative to using babel-polyfill:
[...]

PR #1621, opened 2018-03-23
feat(MdTable): Implement an asynchronous table

emman27:
An initial implementation of a table with data hosted on a JSON endpoint.
- [x] Fetches from endpoint
- [x] Allows using a nested data structure
- [x] Allows custom headers
- [x] Allows custom parameters in the querystring of the request
- [ ] Link up with pagination


package.json
```diff
@@ -41,11 +41,12 @@
   "peerDependencies": {
     "vue": "^2.5.0"
   },
-  "dependencies": {},
+  "dependencies": {
+    "axios": "^0.18.0"
+  },
```
VdustR: why `axios` changed to dependencies?

src/components/MdTable/MdTable.vue
```diff
@@ -123,6 +124,20 @@
...
+      asyncResultsKey: {
+        type: String
+      },
+      asyncParams: {
+        type: Object,
+        default: () => {}
+      },
+      asyncHeaders: {
+        type: Object,
+        default: () => {}
```
VdustR: All `vue-material` props should be started with prefix `md`

src/components/MdTable/MdTable.vue
```diff
@@ -296,6 +312,10 @@
       // wait for `selectingMode` from `TableRow`
       await this.$nextTick()
       this.syncSelectedValue()
+      if (this.asyncUrl) {
+        const data = (await axios.get(this.asyncUrl, { params: this.asyncParams, headers: this.asyncHeaders })).data
```
VdustR: Should there be a loading UI while requesting?
How about a named slot with indeterminate loading bar as default?

package.json
emman27: Could potentially just use native fetch, tho fetch isn't supported in all browsers. Any other preferences

VdustR: I mean all modules imported are in `devDependencies ` for now.

VdustR:
@emman27 check list markdown is `- [ ]`, `- [x]`

src/components/MdTable/MdTable.vue
```diff
@@ -45,6 +45,9 @@
       <slot name="md-table-pagination" />
     </md-content>
 
+    <slot name="md-table-loading"></slot>
```
VdustR: only if asyncLoading

src/components/MdTable/MdTable.vue
```diff
@@ -45,6 +45,9 @@
...
+    <md-progress-bar md-mode="indeterminate" v-if="!$scopedSlots['md-table-loading'] && asyncLoading" />
+
```
VdustR: How about this?

```html
<slot name="md-table-loading" v-if="asyncLoading">
  <md-progress-bar md-mode="indeterminate" />
</slot>
```

package.json
emman27: Hmm. I think if there is going to be to support a http client natively to this project, then the client used internally needs to be either a dependency or a peer dependency. Adding as dev dependency will eventually result in it not being installed by `npm install vue-material`, which may cause confusion for users.

Also open to replacing this with `XMLHttpRequest` (preferrably not, since this has really fallen out) or `fetch` (with a polyfill installed as dependency), depending on your team's preference for direction

src/components/MdTable/MdTable.vue
emman27: Thanks for helping me out with this! Not extremely familiar with slots myself.

package.json
VdustR: You don't need to install the modules which vue-material used because the `vue-material` module you importing is production that already compiled with all modules imported to it.

VdustR: For example, you can import `vue-material` without problem if you don't install `date-fns` and `popper.js`.

src/components/MdTable/MdTable.vue
```diff
@@ -123,6 +129,20 @@
...
+      mdAsyncResultsKey: {
+        type: String
+      },
+      mdAsyncParams: {
+        type: Object,
+        default: () => {}
+      },
+      mdAsyncHeaders: {
+        type: Object,
+        default: () => {}
```
VdustR: type could be `Function`

VdustR:
Hey @emman27, would you resolve conflicts? Actually it's better to start from `dev` branch to make a PR.

VdustR:
You need to fetch current `dev` before you merge that.

emman27:
Thanks, was a careless mistake and pushed too early. Fixed now.

VdustR:
Although this is almost completed, I think make `v-model` like `md-options` as a `Promise` in `MdAutocomplete` might be better for most cases.

Samuell1:
Thanks for your first PR! First i think **axios** should not be a package that we depends on. Best to make it lightweight is to make a callback like `md-autocomplete` is using.

And please dont use `await`, `async` it will broke code - #1182

Samuell1:
@emman27 Thanks you too for contributing and using Vue Material! 

And last thing dont forget in that commits naming `:` after `feat(...)`

src/components/MdTable/MdTable.vue
```diff
@@ -357,13 +342,14 @@
     },
     async created () {
```
Samuell1: I think you forgot to remove `async` here

src/components/MdTable/MdTable.vue
```diff
@@ -153,7 +158,9 @@
           getModel: this.getModel,
           getModelItem: this.getModelItem,
           selectingMode: null
-        }
+        },
+        items: this.value || [],
```
Samuell1: I think this should be as Computed

emman27: Hmm I think I'll disagree with this one. This is because we'll have to set the value once the Promise resolves.
Vue recommends not modifying props directly and instead using a data property, or computed if possible. Since there's a set involved, shouldn't data be the more natural choice?

Samuell1: Yes but when i pass my array it will be not reactive to changes because data are changed only on mounted. 

And there are few things that you forgot to change `2 methods (getModel, getModelItem) that are still pointing to old this.value`

emman27: Hmm good point. I believe this can be resolved by
```
watch: {
  value() {}
}
```
as well. Which do you prefer?

Samuell1: Yeah, watch should be good

src/components/MdTable/MdTable.vue
```diff
@@ -215,6 +215,7 @@
       return { MdTable }
     },
     watch: {
+      value() {}, // Watch this prop to make the table reactive to props
```
Samuell1: This is not finished? You missing here to set value as items.

src/components/MdTable/MdTable.vue
```diff
@@ -215,7 +215,9 @@
       return { MdTable }
     },
     watch: {
-      value() {}, // Watch this prop to make the table reactive to props
+      value() {
+        if (isPromise(this.value)) this.resolvePromise()
```
Samuell1: you need to implement `else` too because all other tables that doesnt use promise are broken.

Samuell1: I think better is to make it as computed

```
items () {
  if (isPromise(this.value)) {
    return this.resolvePromise()
  } else {
    return this.value
  }
}
```

VdustR: There shouldn't be a mutation in a computed.

emman27: Thanks guys. Care to share the secret of how you guys are so sharp?

@Samuell1 I think that example doesn't work because `this.resolvePromise` doesn't (and cannot?) return the resolved value.

Samuell1: @emman27 It will work like `then((resolve) => (resolve))`, but @VdustR said it will be needed to change resolvePromise method.

I think this will now work.

asadhshujau:
when will we get pagination feature. waiting for it ...

dragosct10:
Hi, @emman27! Thanks for your work. Can you please resolve the conflicts to merge this PR?

Issue #1625, opened 2018-03-24
TypeError in MdAutocomplete.vue


maarteNNNN:
### Steps to reproduce
I open this issue to refer to my pull request.
`Md.Autocomplete.vue` has a bug which wants `toLowerCase` a string. But isn't always defined as a string. That's why there should be a `toString().toLowerCase()`. So there no nasty TypeError.

### Which browser?
Chromium Linux v65

### What is expected?
No TypeError

### What is actually happening?
Getting a TypeError `toLowerCase() is not a function` like this:
![2018-03-24_18-57-05_screenshot](https://user-images.githubusercontent.com/14275291/37869443-653a32ec-2f96-11e8-8222-399508a9fbc7.png)

VdustR:
Related to https://github.com/vuematerial/vue-material/issues/1243

Samuell1:
Duplicate of #1243

PR #1626, opened 2018-03-24
updated this.searchTerm.toLowerCase()

maarteNNNN:
Updated to `this.searchTerm.toString().toLowerCase()` to get a TypeError.
Issue #1625


VdustR:
The better way to fix this is make autocomplete supporting objects.

There was a PR for this. #1338

We are trying to find a non-breaking change to fix it.

Samuell1:
Thanks for PR! But this needs to be solved it that other PR like @VdustR pointed.

Issue #1588, opened 2018-03-08
[MdMenu] does not close in safari/iOS


idlinnes:
I used MdMenu in a website, it works with desktop safari, but when I open it in a ios version of safari(IPAD\IOS10-11), the menu does not close when I click outside or clicked one item. The same happened with IE11 but i fix it by adding polyfill to Node.contains.

I can reproduce it with https://www.browserstack.com
Link to video - https://drive.google.com/open?id=1Mye-13daptmoCcst1iizTIvwjvTGutSl

Samuell1:
> this goo.gl shortlink has been disabled. It was found to be violating our Terms of Service. Click here and here for more information about our terms and policies respectively.

Samuell1:
Do you get any errors in console ?

idlinnes:
> Do you get any errors in console ?
no.

andreofthecape:
Have something similar, viewing vuematerial website on osx chrome, safari and firefox the menu's only close when I click outside, but not by clicking on item. In codepen they close normally by clicking outside and by clicking on item. Also tried windows now and getting same behaviour.

Samuell1:
@andreofthecape its because items in menu doesnt have defined any action thats why you cant click on them, if you add `@click` or prop `to` it should allow clicking.

VdustR:
I can fix that if @marcosmoura agree it should be close when plain text item be clicked.

VdustR:
https://material-components-web.appspot.com/menu.html

It's closed while click on the white padding space.

PR #1630, opened 2018-03-26
feat(MdMenu): new props mdCloseOnClick

VdustR:
* new props `:md-close-on-click` option for auto close after any click event.
* `:md-close-on-click` is reactive.
* make `:md-close-on-select` reactive.
* example for props `:md-close-on-select` and `:md-close-on-click`.

~~BREAKING CHANGE: enable auto close after any click event as default~~ (Reverted)

related to https://github.com/vuematerial/vue-material/issues/1588#issuecomment-371747185

### DIFF
--- a/src/components/MdMenu/MdMenu.vue
+++ b/src/components/MdMenu/MdMenu.vue
@@ -30,6 +30,10 @@
         type: Boolean,
         default: true
       },
+      mdCloseOnClick: {
+        type: Boolean,
+        default: true
+      },
       mdSize: {
         type: String,
         default: 'small',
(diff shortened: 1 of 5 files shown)


src/components/MdMenu/MdMenu.vue
```diff
@@ -30,6 +30,10 @@
         type: Boolean,
         default: true
       },
+      mdCloseOnClick: {
+        type: Boolean,
+        default: true
```
marcosmoura: I think this should be false on start. That's because we may have a menu that does not implement lists inside. For some "filter" panels for example.

Issue #1503, opened 2018-02-10
[MdMenu] does not close in chrome/iOS


oeway:
I used MdMenu in a website, it works with desktop chrome, but when I open it in a ios version of chrome, the menu does not close when I click outside or clicked one item. The same happened with Safari/iOS version. I can't reproduce it with the mobile emulator in chrome dev tools. Not sure what happened.

Samuell1:
Can you reproduce it on https://www.browserstack.com ?

wozturk:
I think this issue is unresolved:
To reproduce open https://vuematerial.io/components/menu/ on the iPhone simulator (tested on iOS 10) or on an iPhone 7 with iOS 10, click on any of the options to open a menu (eg the first option "bottom start"). Now click/touch anywhere outside of the menu that just popped up... the menu will not close
Happens on chrome, safari and firefox on iOS.
Btw the menus do close on Android devices

csteeg:
@Samuell1 could you please re-open the issue? It's reproducable on the url @wozturk  mentioned on browserstack as well #os=iOS&os_version=9.0&device=iPhone+6S&device_browser=safari) 
The menus only close when opening other menus, there is no other way to close the menus on iOS

Samuell1:
Maybe its solved here: #1588

wozturk:
@Samuell1 I think #1588 is an issue about the links inside the menu. These seem to work fine when implemented correctly. This issue is about being able to dismiss the popup menu by clicking outside of the box, which isn't working on iOS

Samuell1:
@wpots I missclicked and posted a id to issue not to a PR: this should solve any problems with clicking outside too #1630

PR #1634, opened 2018-03-27
Fix for issue #1503 -> ios devices do not trigger click events on 'no…

csteeg:
iOS doesn't trigger click on 'non-clickable' elements:
https://stackoverflow.com/questions/3705937/document-click-not-working-correctly-on-iphone-jquery/17490775


package-lock.json
```diff
@@ -0,0 +1,18362 @@
...
+        "invariant": "2.2.4",
+        "semver": "5.5.0"
+      }
+    },
+    "babel-preset-flow": {
+      "version": "6.23.0",
+      "resolved": "https://registry.npmjs.org/babel-preset-flow/-/babel-preset-flow-6.23.0.tgz",
+      "integrity": "sha1-5xIYiHCFrpoktb5Baa/7WZgWxJ0=",
+      "dev": true,
+
```
VdustR: We use `yarn` instead of `npm` for Vue Material. Please remove `package-lock.json`.

PR #1635, opened 2018-03-27
md-field @md-clear event

bentatum:


src/components/MdField/MdField.vue
```diff
@@ -47,6 +47,7 @@
       mdDense: Boolean,
       mdNested: Boolean,
       mdClearable: Boolean,
+      mdClear: Function,
```
marcosmoura: This is not needed. Could you revert it?

bentatum:
All set @marcosmoura. Some tests are failing, but it appears unrelated to this change. Let me know if there's anything else I can do.

marcosmoura:
@bentatum Can you add this option to Fields docs page?

bentatum:
@marcosmoura I added an events table with `md-clear` on the MdField docs page.

PR #1637, opened 2018-03-27
nl_BE translation should be good enough for nl_NL aswell.

maarteNNNN:

### DIFF
--- a/docs/app/i18n/en-US/pages.js
+++ b/docs/app/i18n/en-US/pages.js
@@ -91,7 +91,7 @@ export default {
     },
     lastRow: {
       title: 'The last toolbar row fixed',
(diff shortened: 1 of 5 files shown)


maarteNNNN:
They're not finished yet. Need to do some refining. The first commit is discarded, I'm still struggling with how to use git properly. I made another pull request for that #1638

marcosmoura:
@maarteNNNN First of all, thank you so much for starting this branch!
I don't know any about Dutch. This translation works for the Netherlands as well? What's the main difference?

maarteNNNN:
@marcosmoura They're basically the same. They have few words that have a different meaning. Mostly used in dialects. It's not as huge of difference compared Portuguese from Portugal and Brazil
And check your mail marcosvmmoura@gmail.com, I sent you an email ;)

maarteNNNN:
This should be production ready, might be some flaws. There's only two sentences I have difficulty putting into words. Maybe someone else of the community can help.

PR #1645, opened 2018-03-29
feat(MdList): expand only one option

### LINKED ISSUE #1641: [MdList] option to expand only one list item when click
I have some problem with md-expand in new version. In old version, we have a component **`md-list-expand`** which I think is work very well, but for new version, **`md-list-expand`** was removed, and we use attribute **`md-expand`** of **`md-list-item`** instead. With old version, when I click on one **`md-list-expand`** item then it will expand content below and all other expanded item will close, but it not work like this with new version. Is there have any option on new version to work like old version or I need to do myself by remove all md-active class before click to expand on any list item?

VdustR:
## new props `md-expand-only-one`

> If set true, one expandable list item could be expanded at most at a time. The expanded list item will be collapsed when another is expanded.

fix #1641

### DIFF
--- a/docs/app/pages/Components/List/List.vue
+++ b/docs/app/pages/Components/List/List.vue
@@ -56,6 +56,7 @@
       <api-item title="API - md-list">
         <p>The following options can be applied to any list:</p>
 
+        <api-table :headings="list.props.headings" :props="list.props.props" slot="props" />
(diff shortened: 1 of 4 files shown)


marcosmoura:
@VdustR I think a better name for this prop could be `md-expand-single`

Issue #1602, opened 2018-03-14
[MdDatepicker] typing broken as of 1.0.0-beta-8


emman27:
### Reproduction Link

In 1.0.0-beta-7, you could use the keyboard to input a date in the format YYYY-MM-DD.
In 1.0.0-beta-8, this method of input has become extremely clunky with the datepicker aggressively filling up the date.

Random fact: The way the remaining numbers are filled up are different for Chrome and Firefox (the 2 browsers I have on my machine)

(Sorry for the random letters at the end, I just got really frustrated at the datepicker)

![datepicker](https://user-images.githubusercontent.com/6295583/37380614-69e0f2b0-2774-11e8-87e6-042348cd9bbe.gif)

Samuell1:
Its because in beta-7 was enabled native "date" input you can enable it by prop `md-override-native` to false

emman27:
Just another suggestion: I think plaintext is fine, with a debounce on input for the conversion to a Date object, and simply don't do the conversion if the string isn't a valid date.

Samuell1:
@emman27 good idea! I loot at it later.

qubitron:
Being able to type in dates is very important for data entry, so glad you are looking into this! When addressing this you may need to take into account the dateformat used in the current locale, similar to  #1422

Samuell1:
@qubitron you can change format of date with `this.$material.locale.dateFormat = 'YYYY-MM-DD'`

qubitron:
@Samuell1 I wasn't having trouble with setting the date format. I was just trying to be helpful and suggesting the date format should also be used when parsing date input as well, since the built-in JavaScript date type may not accept many date locales as valid.

Samuell1:
@qubitron oh sorry i misunderstand it :D yeah thats true it should be both ways

PR #1666, opened 2018-04-04
feat(MdDatepicker): Debounce the conversion into a date object

emman27:
Fixes https://github.com/vuematerial/vue-material/issues/1602

Random note: Some tests failing on other components - do I have to fix these? Seems all the PRs are affected

### DIFF
--- a/src/components/MdDatepicker/MdDatepicker.vue
+++ b/src/components/MdDatepicker/MdDatepicker.vue
@@ -28,7 +28,7 @@
   import MdOverlay from 'components/MdOverlay/MdOverlay'
   import MdDatepickerDialog from './MdDatepickerDialog'
   import MdDateIcon from 'core/icons/MdDateIcon'
-  import debounce from 'core/utils/MdDebounce'
+  import mdDebounce from 'core/utils/MdDebounce'
   import MdField from 'components/MdField/MdField'
   import MdInput from 'components/MdField/MdInput/MdInput'
 
@@ -56,6 +56,10 @@
         type: Boolean,
         default: false
       },
+      mdDebounce: {
+        type: Number,
+        default: 1000
+      }
     },
     data: () => ({
       showDialog: false,
@@ -94,12 +98,12 @@
       }
     },
     methods: {
-      onInput: debounce(function (value) {
+      onInput(value) {
         const parsedDate = parse(value)
         if (isValid(parsedDate)) {
           this.selectedDate = parsedDate
         }
-      }, 1000),
+      },
       toggleDialog () {
         if (!isFirefox || this.mdOverrideNative) {
           this.showDialog = !this.showDialog
@@ -133,6 +137,7 @@
       }
     },
     created () {
+      this.onInput = mdDebounce(this.onInput, this.mdDebounce)
       this.modelDate = this.dateToHTMLString(this.value)
       this.selectedDate = this.value
     }
--- a/src/components/MdDatepicker/MdDatepicker.vue
+++ b/src/components/MdDatepicker/MdDatepicker.vue
@@ -1,7 +1,7 @@
 <template>
   <md-field :class="['md-datepicker', { 'md-native': !this.mdOverrideNative }]" md-clearable>
     <md-date-icon class="md-date-icon" @click.native="toggleDialog" />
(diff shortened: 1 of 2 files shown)


Samuell1:
Hey, you can use our core util `debounce` function https://github.com/vuematerial/vue-material/blob/dev/src/core/utils/MdDebounce.js

src/components/MdDatepicker/MdDatepicker.vue
```diff
@@ -93,6 +94,12 @@
       }
     },
     methods: {
+      onInput: debounce(function (value) {
+        const parsedDate = parse(value)
+        if (isValid(parsedDate)) {
+          this.selectedDate = parsedDate
+        }
+      }, 1000),
```
VdustR: Debounce should be used in this way to make scope in the instance instead of in the component.

```javascript
export default {
  // ...

  methods: {
    onInput (value) {
      const parsedDate = parse(value)
      if (isValid(parsedDate)) {
        this.selectedDate = parsedDate
      }
    }
  }

  created () {
    this.onInput = debounce(this.onInput, 1000)
  }

  // ...
}
```

You could try this example to know the difference.

```vue
<template>
  <div>
    <debounce-test></debounce-test>
    <debounce-test></debounce-test>
  </div>
</template>

<script>
  import MdDebounce from 'core/utils/MdDebounce'

  export default {
    name: 'Debounce scope compare',

    components: {
      debounceTest: {
        name: 'DebounceTest',

        data () {
          return {
            initTime: null,
            wrongDebounceResult: [],
            rightDebounceResult: []
          }
        },

        methods: {
          wrongDebounce: MdDebounce(function () {
            this.wrongDebounceResult.push(new Date() - this.initTime)
          }, 1000),

          rightDebounce () {
            this.rightDebounceResult.push(new Date() - this.initTime)
          },

          exec () {
            this.wrongDebounce()
            this.rightDebounce()
          }
        },

        render (createElement) {
          return createElement('div', 
            [
              createElement(
                'div', `wrong: ${this.wrongDebounceResult}`
              ),
              createElement(
                'div', `right: ${this.rightDebounceResult}`
              )
            ]
          )
        },

        created () {
          this.rightDebounce = MdDebounce(this.rightDebounce, 1000)
          this.initTime = new Date()

          setTimeout(this.exec(), 1000)
        },
      }
    }
  }
</script>
```

Result:
```
wrong:
right: 1345
wrong: 1360
right: 1371
```

emman27: Hey!

I'm all for your implementation - I also see that this allows the debounce time to be set as a prop rather than hardcoded at 1000ms.

Would you mind explaining why the results are wrong though? I don't quite understand your example.

src/components/MdDatepicker/MdDatepicker.vue
```diff
@@ -28,25 +28,10 @@
   import MdOverlay from 'components/MdOverlay/MdOverlay'
   import MdDatepickerDialog from './MdDatepickerDialog'
   import MdDateIcon from 'core/icons/MdDateIcon'
+  import debounce from 'core/utils/MdDebounce'
```
Samuell1: It should be renamed from `debounce` to `MdDebounce` because it can later conflict with other libraries.

src/components/MdDatepicker/MdDatepicker.vue
VdustR: As I said scope shouldn't in the compnent, it should be in the instance. The wrong one has side effect that the second instance use the same `timeout` with the first one. It is expected to execute the method after the last call with debounce per instance. There shouldn't be side effect cross the instances.
For example, if I use a longer delay, or my device performance is bad, or I focus and type in another datepicker quickly, the previous datepicker `onInput` will not be called.

emman27:
Thanks guys!
@Samuell1 I've gone with `mdDebounce` instead to keep the naming convention of functions being `camelCase`.

VdustR:
Would you update the documentation with the new prop?

VdustR:
Sorry, I missed. Thanks.

Samuell1:
@emman27 function from **core/utils** should be named with first uppercase because we using props with camelcase and it can conflict.

PR #1698, opened 2018-04-28
refactor(MdDatepicker): Better flow

### LINKED ISSUE #1689: [MdDatepicker] using format 'DD/MM/YYYY' exceeding stack size
### Steps to reproduce
Setting `Vue.material.locale.dateFormat = 'DD/MM/YYYY'`.
Trying to select a date using `md-datepicker`.

### Which browser?
Chrome Canary.

### What is expected?
Be able to select a date.

When selecting a date using the dateFormat is `DD/MM/YYYY` the watcher for `selecteDate` is again executed and then there is loop.

https://github.com/vuematerial/vue-material/blob/dev/src/components/MdDatepicker/MdDatepicker.vue#L88

### Reproduction Link
https://codesandbox.io/s/6398yyjqn

VdustR:
- Upgrading `date-fns` to `2.0.0-alpha.7`
    For using new parser API because we have custom date string format
- Better naming and structure for `inputDate`(the string from input), `localDate` and `value`(from `v-model`)
- New feature: `v-model` supports `number` and `string` now, and its type is reactively modifiable via changing the type of `v-model` or changing the new props `:md-model-type`.
    Including documentation
- Debounce for input (will be conflicted with #1666. I'll add that props after it merged)

BREAKING CHANGE: upgrade `date-fns` to `2.0.0-alpha.7`

fix #1689


marcosmoura:
@VdustR I think this is not good with the HTML5 standard unless we provide a "format" prop. To pass something like "dd/mm/yyyy". Do you think you can implement something like that?

marcosmoura:
I mean, we should document the locale.dateFormat.

VdustR:
I added date format option into basic example. Additionally, I think there could be another page to explain all vue material options. How do you think?

docs/app/pages/Configuration.vue
```diff
@@ -0,0 +1,78 @@
...
+          locale: {
+            ...Vue.material.locale,
+            dateFormat: 'DD/MM/YYYY',
+            firstDayOfAWeek: 1
+          }
+        }
+      </code-example>
+      <p>or you can change it via <code>this.$material</code> in a vue component:</p>
+      <code-example :title="$t('pages.configuration.inVueComponents')">
+        export default {
```
marcosmoura: This can also be a table with the API. What do you think?

VdustR: I don't know. The configuration object is nested. I'm afraid that table can't describe its structure clearly.

hastom:
@VdustR @Samuell1 
This PR breaks datepicker completely. 
I think as it was early alpha date-fns formats were just buggy, so behavior of datepicker on current dev branch looks like this:
image
Also date-fns date formats changed after alpha.7

PR #1731, opened 2018-05-14
fix(MdTable): fix double default slots

Samuell1:
Simple table without v-model is rendering default slots 2 times.
Reproduction: https://codesandbox.io/s/1qy64o65j?module=App.vue


fabioviana-hotmart:
This problem still occurs when it has the component md-table-pagination.

Issue #1680, opened 2018-04-12
[MdButton] Stop using `<div>` within `<button>`


kdex:
Version: `1.0.0-beta-8`

After rendering, an `md-button` will become an element of type `<button>`.

But if the `md-button` uses a ripple animation (which in turn uses `<div>` elements), the end result will be a `<button>` containing a `<div>`, which is invalid HTML markup.

Errors like these could easily be checked for pre-rendered sites by [validating](https://html5.validator.nu/) the HTML.

VdustR:
related to #1461

Samuell1:
@kdex from latest version beta-10 ripple now is using spans to generate ripple waves, but still ripple parent is still div

marcosmoura:
The HTML5 validation is not a thing to consider these days. The div inside button is only because Firefox can't work with `display: flex` for buttons. So that's why we need that.

kdex:
> The HTML5 validation is not a thing to consider these days.

@marcosmoura Please do correct me if I'm misinterpreting you, but did you mean to say that invalid markup is acceptable these days? If so, I couldn't disagree more. Markup bears implicit semantics that are important for screen readers, search engines and UX in general. Just because often times, "it renders just fine" doesn't mean it's justified to use it, as the visual rendering is just one aspect of what markup describes.

There's much more to it than what's visible on your screen.

Would you please re-open this issue until it's fixed?

marcosmoura:
@kdex What I meant to say is that we shouldn't be that focused on the strict markup validation. Semantics and right markup is a must for several reasons, but I believe that there is more than just semantics to think here. 

Just because the button element doesn't "allow" divs inside does not mean that is not fine to use it. In fact we have to use not only for ripple effects but also to support other elements than just text inside buttons. If we think on Chrome it is OK to have only the button with a bunch of other inline children. But we also have to support IE, Safari, Firefox, UC Browser and many more. That's why we also have to use some other quirks to achieve what we want.

Regarding screen readers, in my personal tests, the buttons didn't have a problem with the text recognition and text to speech.

kdex:
@marcosmoura If the problem is related to Firefox's `display: flex` behavior, is it out of the question to implement a solution using methods that predate flexboxes?

marcosmoura:
Yep. It is.
For example the `md-tabs` and `md-list-item` components. They have a lot of flexibility that Flexbox solves for us.

[...]

PR #1744, opened 2018-05-15
docs: corrections of breakpoints

### LINKED ISSUE #1653: [MdListItem] Clickable MdListItem hides its action button's click
### Steps to reproduce

1. Create a list containing a `<md-list-item/>`.
2. Add a click handler to that list item: `<md-list-item @click="console.log('clicked')"/>`.
3. Add a button as an action to that item:
   ```js
   <md-list-item @click="console.log('clicked')">
     <md-button class="md-list-action" @click="console.log('action')"/>
   </md-list-item>
   ```
4. Try to click the list item's action button.

### Which browser?

Firefox 59.0.2
(it *works* in Chrome 65.0.3325.181)

### What is expected?

The button reacts to hover and triggers its click action.

Only the list item's "selection" is triggered. The button displays neither a hover effect, nor is its click action called.

### Reproduction Link

https://codesandbox.io/s/p4y4ykxo7

### LINKED ISSUE #1685: Corrections to Breakpoints in online docs
This is not a bug report for the framework, but for the online docs regarding the breakpoints table here:

https://vuematerial.io/ui-elements/layout

Assuming the breakpoints work similarly to Bootstrap, then the following corrections should be made:

- **xsmall**:  "maximum of 600px" -> "maximum of 599px"
- **small**:  "between of 600px and 960px" -> "from 600px to 959px"
- **medium**: "between of 960px and 1280px" -> "from 960px to 1279px"
- **large**: "between of 1280px and 1920px" -> "from 1280px to 1919px"
- **xlarge**: "bigger than 1920px" -> "1920px and larger"

This makes the breakpoint boundaries very clear (e.g., in its current form it isn't entirely clear whether 600px belongs to "xsmall" or "small" breakpoint).

Samuell1:
Resolves #1685

### DIFF
--- a/docs/app/pages/Components/List/List.vue
+++ b/docs/app/pages/Components/List/List.vue
@@ -39,11 +39,6 @@
         <li v-for="event in interactionEvents" :key="event">{{ event }}</li>
       </ul>
 
-      <note-block>
-        Mixing a list action with actions on the list items hurts usability. When clicking on the list item action, will the list action be triggered as well?
-        Therefore, when using a list action, do not use actions on the list items. Mixing them additionally has technical compatibility issues between browsers.
-      </note-block>
-
       <p>If you pass a <code>href</code> attribute, the item will automatically render an <code>a</code> tag.</p>
       <p>All <a href="https://router.vuejs.org/en/api/router-link.html" target="_blank">options</a> of <code>router-link</code> can be simply used here. This will render an <code>a</code> tag:</p>
       <code-example title="Different types" :component="examples['list-types']" />
--- a/docs/app/pages/Components/List/List.vue
+++ b/docs/app/pages/Components/List/List.vue
@@ -39,6 +39,11 @@
         <li v-for="event in interactionEvents" :key="event">{{ event }}</li>
       </ul>
 
+      <note-block>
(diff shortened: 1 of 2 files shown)


docs/app/pages/Components/List/List.vue
```diff
@@ -39,6 +39,11 @@
         <li v-for="event in interactionEvents" :key="event">{{ event }}</li>
       </ul>
 
+      <note-block>
+        Mixing a list action with actions on the list items hurts usability. When clicking on the list item action, will the list action be triggered as well?
```
marcosmoura: I found this tip a little confusing. Maybe we can have more examples to show why this will happen. What do you think?

Samuell1: I am not sure, maybe better will be to link to issue?

VdustR: Related to #1653.

Probable solution #1680.

Samuell1: @VdustR yes i know but i mean linking in docs to that issues will be better then examples

VdustR: I agree about `<div>` if there is no better way to implement that.

PR #1769, opened 2018-05-22
Resolve #1677 (infinite file picker loop)

### LINKED ISSUE #1677: [MdFile] Clicking on the file input field in NW.js triggers endless file selection dialog box
### Steps to reproduce
Download the latest NW.js for Windows, added a simple HTML file to test the File Input Field of Vue Material. Click on the File Input Field.

### Which browser?
NW.js 64 bit on Windows 10

### What is expected?
Selecting a file or canceling the selection will close the file selection dialog box

The file selection dialog box will keep coming up no matter what you do. The only thing that works is to kill the app.

0xADD1E:
Using `@focus` on md-file would trigger an infinite loop of the file picker by re-sending the focus event as soon as the file picker closed.

Resolves #1677

### DIFF
--- a/src/components/MdField/MdFile/MdFile.vue
+++ b/src/components/MdField/MdFile/MdFile.vue
@@ -7,7 +7,7 @@
       readonly
       v-model="model"
       v-bind="{ disabled, required, placeholder }"
-      @focus="openPicker"
+      @click="openPicker"
       @blur="onBlur">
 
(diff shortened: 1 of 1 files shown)


Samuell1:
@0xADD1E thanks for fix, its compatible for all browsers?

0xADD1E:
Tested on current firefox, chrome canary, safari

PR #1775, opened 2018-05-23
fix(MdTable): sort on created

### LINKED ISSUE #1774: [MdTable] md-sort not applied on initial render
Given the following example: https://vuematerial.io/components/table/

Scroll down to the table sample for Sort, and the md-sort property is set to Name, as indicated by the arrow adjacent to the Name column header.

The data however is unsorted (the sequence being IDs 1,2,3,4,5 as in the underlying dataset). Click the column header to actually apply a sort by Name.

Thus the md-sort property is only enabling the sort facility, but on first load, whilst the directional arrow is drawn to indicate that the data is being sorted, the data is not actually sorted until the column header is clicked, rather it remains unsorted and is displayed in the same sequence as it exists in the underlying array.

Screenshot for clarity.

screen_shot_2018-05-23_at_13_46_27

This is using 1.0.0-beta-10.2.

This is my first time raising an issue, apologies if I have got anything wrong and my huge thanks to all involved for their great work.

VdustR:
fix #1774

### DIFF
--- a/src/components/MdTable/MdTable.vue
+++ b/src/components/MdTable/MdTable.vue
@@ -348,6 +348,10 @@
       }
     },
     created () {
+      if (this.mdSort) {
+        this.sortTable()
+      }
+
       this.$nextTick().then(() => {
         this.syncSelectedValue()
       })


ahmadshahrour951:
Hey @Samuell1, do you know when will this be live in production? >.<

Samuell1:
@ahmadshahrour951 Hey, it should be already in stable.

ahmadshahrour951:
thank you 👍

Issue #1670, opened 2018-04-07
Multiple theme registration always output the first registered theme name in md-base-theme() mixin


simongcc:
### Steps to reproduce

Because my layout have several button and page design.
So the base code look like this:
```
@include md-register-theme("default", (
  primary: md-get-palette-color(grey, 700),
  secondary: md-get-palette-color(black, A200),
  accent: md-get-palette-color(white, A200),
  // theme: light // default
));

@include md-register-theme("dark-solid", (
  primary: md-get-palette-color(black, A700),
  accent: md-get-palette-color(blue, A200),
  // primary: md-get-palette-color(white, A700), // The primary color of your application
  // accent: md-get-palette-color(red, A200), // The accent or secondary color
));
```
### What is expected?

It is expected the theme to be separately generated.
So that each theme can work in harmony.
```
:root {
  --md-theme-default-primary: #616161;
  --md-theme-default-accent: #fff;
  --md-theme-default-theme: "light";
  --md-theme-default-secondary: #000;
}
.md-theme-default :not(input):not(textarea)::-moz-selection {
  background-color: #fff;
  background-color: var(--md-theme-default-accent-on-background, #fff);
  color: rgba(0, 0, 0, 0.87);
  color: var(--md-theme-default-text-primary-on-accent, rgba(0, 0, 0, 0.87));
}
... skip

:root {
  --md-theme-dark-solid-primary: #000;
  --md-theme-dark-solid-accent: #448aff;
  --md-theme-dark-solid-theme: "light";
}
.md-theme-dark-solid :not(input):not(textarea)::-moz-selection {
  background-color: #fff;
  background-color: var(--md-theme-default-accent-on-background, #fff);
  color: rgba(0, 0, 0, 0.87);
  color: var(--md-theme-default-text-primary-on-accent, rgba(0, 0, 0, 0.87));
}
/* ... skip */
```

### What is actually happening?

The outcome css look like this:
As you can see, the non-default theme is also output as "md-theme-default" in md-base-theme() mixin.
```
:root {
  --md-theme-default-primary: #616161;
  --md-theme-default-accent: #fff;
  --md-theme-default-theme: "light";
  --md-theme-default-secondary: #000;
}
.md-theme-default :not(input):not(textarea)::-moz-selection {
  background-color: #fff;
  background-color: var(--md-theme-default-accent-on-background, #fff);
  color: rgba(0, 0, 0, 0.87);
[...]

PR #1783, opened 2018-05-29
Fix the multiple theme registration generate duplicated set of CSS classes

simongcc:
Refer to issue #1670
Fix the duplication in generating multiple css classes when there is multiple themes by adding a `$md-theme-name` variable to track the current reading theme during running the css generation loop while generating all css for all components.
It fix the following mixins
`md-theme-component()`
`md-theme-component-relative()`


simongcc:
I think I need to update the commit message, because when I committed before, I didn't thought the solutions will be accepted as PR.

PR #1796, opened 2018-06-02
Fix path to MdTheme

DavidHavl:
Fix path to MdTheme since it is in same folder

### DIFF
--- a/src/core/MdComponent.js
+++ b/src/core/MdComponent.js
@@ -1,4 +1,4 @@
-import MdTheme from 'core/MdTheme'
+import MdTheme from './MdTheme'
 import deepmerge from 'deepmerge'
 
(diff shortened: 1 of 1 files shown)


src/core/MdComponent.js
```diff
@@ -1,4 +1,4 @@
-import MdTheme from 'core/MdTheme'
+import MdTheme from './MdTheme'
```
VdustR: That's alias
https://github.com/vuematerial/vue-material/blob/44b0147c5c9d4c541046c0b06f99d055ad13a5a5/build/config.js#L36

DavidHavl: Yes but when one wants to use the component on it's own (importing it when creating new components outside) it is throwing error.

PR #1816, opened 2018-06-14
fix(MdStep): set tabindex for not active step

### LINKED ISSUE #1570: [Improvement: MdStepper] should not be accessible if step is not active
Expecting when pressing `Tab` key on last input of a step to highlight the next steps only.

It does highlight the steps, but also the inputs inside it.

Example (assume 3-steps stepper):

- Active step is `step 1` and focusing on last input
- Press `Tab`
- Highlights `step 2` which has 3 inputs
- Press `Tab` expects to highlight `step 3` but instead need to `Tab` 3 times (because has 3 inputs) before highlighting `step 3`

Samuell1:
Resolves #1570 

This will solve issue but removes a transform animation from vertical step.

### DIFF
--- a/src/components/MdSteppers/MdStep.vue
+++ b/src/components/MdSteppers/MdStep.vue
@@ -2,7 +2,7 @@
   <div class="md-stepper">
     <md-step-header v-if="MdSteppers.isVertical" :index="id" />
 
-    <div class="md-stepper-content" :class="{ 'md-active': id === MdSteppers.activeStep }">
+    <div :class="['md-stepper-content', { 'md-active': id === MdSteppers.activeStep }]" :tabindex="tabIndex">
       <slot />
     </div>
   </div>
@@ -40,6 +40,13 @@
         handler () {
           this.setStepperData()
         }
+      }
+    },
+    computed: {
+      tabIndex () {
+        return this.id !== this.MdSteppers.activeStep
+        ? -1
+        : false
       }
     },
     methods: {
--- a/src/components/MdSteppers/MdStep.vue
+++ b/src/components/MdSteppers/MdStep.vue
@@ -2,7 +2,7 @@
   <div class="md-stepper">
     <md-step-header v-if="MdSteppers.isVertical" :index="id" />
 
-    <div :class="['md-stepper-content', { 'md-active': id === MdSteppers.activeStep }]" :tabindex="tabIndex">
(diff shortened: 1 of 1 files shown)


src/components/MdSteppers/MdStep.vue
```diff
@@ -2,7 +2,7 @@
   <div class="md-stepper">
     <md-step-header v-if="MdSteppers.isVertical" :index="id" />
 
-    <div class="md-stepper-content" :class="{ 'md-active': id === MdSteppers.activeStep }">
+    <div :class="['md-stepper-content', { 'md-active': isActive }]" :tabindex="tabIndex" v-show="isActive">
```
VdustR: what is `:tabindex` for?

src/components/MdSteppers/MdStep.vue
```diff
@@ -42,6 +42,16 @@
...
       }
     },
+    computed: {
+      isActive () {
+        return this.id === this.MdSteppers.activeStep
+      },
+      tabIndex () {
+        return !this.isActive
+          ? -1
+          : false
```
xxRockOnxx: Should `false` be `0` instead? Pretty sure that should be a number.

Samuell1: False is because it should not be applied to element.

xxRockOnxx: Just following the HTML specs. Also for confirmation, made a quick Google search: https://stackoverflow.com/a/47322194

Samuell1: Yes but binded tabindex is not a same as html attribute.

VdustR: @xxRockOnxx you can check this with devtool
https://codepen.io/VdustR/pen/EpNGzy

Pass `null`, `false`, `undefined` to attribute on native element node will unset it.

xxRockOnxx: @VdustR I see. Is that because of Vue or not? Because if not I think it's up to the browser what to do with it and that might cause inconsistencies.

VdustR: It's wrote in the documentation.

https://vuejs.org/v2/guide/syntax.html#Attributes

```
If isButtonDisabled has the value of null, undefined, or false, the disabled attribute will not even be included in the rendered <button> element.
```

oliver-ni:
see #2215

Issue #1835, opened 2018-06-28
md-input causes v-model lag on mobile


rayfoss:
### Version
vue 2.5.16
vue-material 1.0.0-beta-10.2

### Reproduction link
[https://codesandbox.io/s/xk52mqm7o](https://codesandbox.io/s/xk52mqm7o)

### Steps to reproduce
type two words on Chrome 66  on an Android phone or emulator. The pixel and pixel images work.

### What is expected?
When you click "create todo" validateForm reads form.todo. Unfortunately, it only updates after every space bar press.

### What is actually happening?
v-model contains everything but the last word

---
I tried v-model.lazy and initializing form to `form: todo: ''`... same issue

![kapture 2018-06-28 at 13 54 57](https://user-images.githubusercontent.com/876076/42055778-bece1a16-7add-11e8-91cf-2a325bb4a7e4.gif)

rayfoss:
@Samuell1 not surprising... disabling spell check makes the problem unnoticeable. This is a race condition. 

My first guess is @click is processed before nextTick, but that doesn't explain why the `<input>` works. <MaterialInput> brute force fixes the issue. Any ideas what I should be looking at? My first inclination is to try to get rid of the watch's and try to use the property more directly

rayfoss:
Good news and bad news:
![screen shot 2018-06-29 at 2 55 34 pm](https://user-images.githubusercontent.com/876076/42112154-98006aea-7bac-11e8-89df-0ba45a3967d7.png)
![screen shot 2018-06-29 at 2 55 25 pm](https://user-images.githubusercontent.com/876076/42112155-98194894-7bac-11e8-9af0-b4af407eb5fc.png)

Samuell1:
@rayfoss can you define your v-model in data? because form.todo doesnt not exists in your object, that is maybe causing issue

rayfoss:
@Samuell1 Yes, I tried that... turns out the setter for the form will handle simple values without the need for an additional per value setter, at least on 2.5.

The best solution I've found so far is to do change the MdInput template to 
```
<template>
  <input
    class="md-input"
    v-model="model"
    v-bind="attributes"
    v-on="listeners"
    @input="$emit('input', $event.target.value)"
    @focus="onFocus"
    @blur="onBlur">
</template>
```

[...]

PR #1836, opened 2018-06-28
Use vue 2 promise syntax for nextTick on MdFields

rayfoss:
https://github.com/vuematerial/vue-material/issues/1835

https://vuejs.org/v2/api/#Vue-nextTick


Samuell1:
Hey, 
callback should be same and it should work, because later they only added support for promise for nextTick.

rayfoss:
doesn't fix the issue and deprecates versions under 2.1

rayfoss:
I have no clue what's causing it... it seems we do a lot of juggling of the model... somewhere in between the watch and setter methods we break phone compatibility compared to a standard input field.

PR #1840, opened 2018-06-29
update to jest 23 and fix moduleFileExtensions

rayfoss:
In base there is `index.scss` which won't be picked up unless configured when using jest 23.


rayfoss:
If you update to jest 23, and try to run tests, it will fail because of the `base/index.scss` file not being whitelisted in the extensions. I had some issues triggering the debugger, to find solutions it helps to be in the latest version...

PR #1946, opened 2018-10-08
fix(MdProgressSpinner): improve transition

### LINKED ISSUE #1935: [MdProgress] Spinner causes delayed v-show DOM cleanup
# Description
A change of state which causes `v-if` or `v-show` to hide the spinner causes DOM cleanup lag of about 1 second.

# Reproduction
https://codesandbox.io/s/qk9lz7452q
Browser: Chrome 69
OS: Linux and Mac

# Proposed Solution
Wrap the spinner in a div

supachailllpay:
### Add explicit transition duration

It seems Vue cannot detect animation/transition end because of infinite animation at https://github.com/vuematerial/vue-material/blob/dev/src/components/MdProgress/MdProgressSpinner/MdProgressSpinner.vue#L260.

### Split enter and leave animation in determinate mode

Use reverse animation for leave state instead.

Fix #1935

### DIFF
--- a/src/components/MdProgress/MdProgressSpinner/MdProgressSpinner.vue
+++ b/src/components/MdProgress/MdProgressSpinner/MdProgressSpinner.vue
@@ -1,5 +1,5 @@
 <template>
-  <transition name="md-progress-spinner" :duration="{ determinate: 2000, indeterminate: 400 }[mdMode]" appear>
+  <transition name="md-progress-spinner" appear>
     <div class="md-progress-spinner" :class="[progressClasses, $mdActiveTheme]">
       <svg
         class="md-progress-spinner-draw"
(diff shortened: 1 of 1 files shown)


supachailllpay:
I think this is broken since the previous PR. I'm not 100% sure but If I understand the code correctly.

The `mdValue` is only used in `circleStrokeDashOffset` which only used in `attachCircleStyle` which used once in `mounted` function. It means nothing will happen if `mdValue` changed.

If I add a watcher the same as `mdDiameter` for `mdValue`. It works fine. 

What do you think?

Samuell1:
Yeah you are true, its broken before too. Yes adding watcher for mdValue should fix that.

supachailllpay:
I have updated my PR.

- Add watcher for `mdValue` and also `mdStroke`.

- Use CSS transition-duration property instead of explicit variable (I think this is more simple).

PR #1957, opened 2018-10-24
fix(MdApp): remove height 100 to make md-app-content span content height

xxRockOnxx:
Title says it all. You can see an example of the problem on the docs page.

https://vuematerial.io/components/app

Check the Fixed Toolbar. After scrolling down, you can see that the border-left of app-content is being left when scrolling.

You can also check the dev tools. The app-content is not the content's height but rather the container's


oliver-ni:
(CI fails but was failing already, not because of PR)

PR #1967, opened 2018-11-11
Fix(MdDatepicker): Propagate clear event

shiragold:
Currently, it is not possible to catch the md-clear event, that comes from MdField and is not propagated by MdDatepicker. Therefore, I added a handler for 'md-clear', which emits the same event.

Currently, the MdDatepicker is always clearable. I added a prop to allow changing this, while 'true' is still the default.

### DIFF
--- a/src/components/MdDatepicker/MdDatepicker.vue
+++ b/src/components/MdDatepicker/MdDatepicker.vue
@@ -1,5 +1,5 @@
 <template>
-  <md-field :class="['md-datepicker', { 'md-native': !this.mdOverrideNative }]" md-clearable>
+  <md-field :class="['md-datepicker', { 'md-native': !this.mdOverrideNative }]" :md-clearable="mdClearable" @md-clear="onClear">
     <md-date-icon class="md-date-icon" @click.native="toggleDialog" />
     <md-input :type="type" ref="input" v-model="inputDate" @focus.native="onFocus" :pattern="pattern" />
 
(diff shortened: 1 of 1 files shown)


Thyiad:
Yes, the datepicker doest really clear the value, just clear the input localValue. 
And I can't find where it did the clear input work..
Very confusing !

src/components/MdDatepicker/MdDatepicker.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <md-field :class="['md-datepicker', { 'md-native': !this.mdOverrideNative }]" md-clearable>
+  <md-field :class="['md-datepicker', { 'md-native': !this.mdOverrideNative }]" :md-clearable="mdClearable" @md-clear="onClear">
```
saxxi: What about

```
<md-field
  :class="['md-datepicker', { 'md-native': !this.mdOverrideNative }]"
  v-bind="$props"
  v-on="listeners">
```

Then add a computed

```
    computed: {
      listeners () {
        return {
          ...this.$listeners,
          // here you can add override listeners
        }
      },
    },
```

gierschv: +1 for this solution!

oliver-ni: I think the current way is fine, to stay consistent with the existing api

PR #1968, opened 2018-11-11
feat(MdInput): add 'isLazy' option

LarenDorr:
When I use md-input components, I try use `v-model.lazy` instead of `v-model`. Because non lazy input that emit more useless events. But that failed.
So, I add `isLazy` prop that determine `md-input` whether to use `v-model.lazy`.


LarenDorr:
Em, Why the lazy input is normal in docs. But When I build and import to my project, that dones't work :(

marcosmoura:
I like this solution, but for now, the code makes the input lazy all the time. The prop should also be prefixed with `md-`

PR #1981, opened 2018-12-07
make label resizing animation smooth

michitaro:
As this [video](https://www.youtube.com/watch?v=2LHkDakbFNA) shows, labels in MDField get wiggling in resizing animation on Safari (including ones in iOS).

Css ```transtion-property: font-size;``` causes wiggling animation of text resizing in Safari, as discussed [here](https://stackoverflow.com/questions/34717492/css-transition-font-size-avoid-jittering-wiggling),

This PR is to use css property ```transform``` instead of ```font-size``` for the labels.
With this change, the resizing animation gets smooth even on Safari.

### DIFF
--- a/src/components/MdField/MdField.vue
+++ b/src/components/MdField/MdField.vue
@@ -170,6 +170,7 @@
       pointer-events: none;
       transition: $md-transition-stand;
       transition-duration: .3s;
+      transform-origin: left;
       font-size: 16px;
       line-height: 20px;
     }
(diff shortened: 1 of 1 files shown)


Samuell1:
@michitaro Hi, thanks for the PR! Is this change is tested for all browsers?

ctexxx:
@Samuell1 The Readme says "for all modern browsers", but what specific browsers are meant by that?

Samuell1:
@ctexxx if you click on that link in readme there is all browsers that are latest and most popular

ctexxx:
Sorry, I didn't see that 😣

PR #1982, opened 2018-12-14
For Internet Explorer 11.

Vadim1Korniychuk:
values not supported


aburai:
i think it is better to use a polyfill, if you use babel try @babel/polyfill or use this simple polyfill

`Object.values = (obj) => Object.keys(obj).map(key => obj[key])`

Samuell1:
Like @aburai said to support IE is best to use polyfills.

PR #2004, opened 2019-01-26
Component/md table pagination fix and docs

saxxi:
Tried to improve @597796340 work on `md-table-pagination` by adding documentation, examples and fixing the following:
- Always notify the caller about pagination changes (regardless pagination is remote or local)
- Set the `isExternalPagination` flag as computed


src/components/MdTable/MdTablePagination.vue
```diff
@@ -4,86 +4,133 @@
...
       <md-field>
-        <md-select v-model="currentPageSize" md-dense md-class="md-pagination-select" @changed="setPageSize">
+        <md-select v-model="currentPageSize" md-dense md-class="md-pagination-select">
           <md-option v-for="amount in mdPageOptions" :key="amount" :value="amount">{{ amount }}</md-option>
         </md-select>
       </md-field>
     </template>
 
-    <span>{{ currentItemCount }}-{{ currentPageCount }} {{ mdSeparator }} {{ mdTotal }}</span>
+    <span>{{ mdPage }}/{{ pageCount }}</span>
```
oliver-ni, line 13: Any reason why this was changed? Should be following material design spec and display the current range.

oliver-ni:
Something like

```js
{{ (mdPage - 1) * currentPageSize }}–{{ mdPage * currentPageSize }} of {{ mdCount }}
```

probably.

oliver-ni:
see #2210

PR #2009, opened 2019-01-30
test: fix test selectors for checkbox and switch

drhaliburton:
Test suite is failing due to label selectors not triggering the proper events for MdCheckbox and MdSwitch tests. Converted tests to use container selectors found in other functioning tests within the same files.

Hopefully this helps with the backlog of PRs that are all failing!

### DIFF
--- a/src/components/MdCheckbox/MdCheckbox.test.js
+++ b/src/components/MdCheckbox/MdCheckbox.test.js
@@ -113,12 +113,12 @@ test('should add and remove a value from model when model is an array by clickin
       model: []
     }
   })
-  const label = wrapper.find('label')[0]
+  const container = wrapper.find('.md-checkbox-container')[0]
 
-  label.trigger('click')
+  container.trigger('click')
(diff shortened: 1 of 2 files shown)


DavidHavl:
Unfortunately I don't think this is the right way to go about it. The tests specifically need to test if a click on LABEL works properly. The changes made are clicking on the container of the input not on a label, so it's not really a point of the test.

DavidHavl:
I have found the issue... it bubbles up and creates double click for some reason.
Here is how to fix it (I am lazy to create a PR):

put everything the way it was (selecting label and clicking on it), but prevent bubbling up.
So instead of 
`label.trigger('click')`

it needs to be: 
`label.trigger('click', {
    cancelBubble: true
  })` 

which basically adds info to the underlying Event created by test script which prevents bubbling up.

DavidHavl:
@marcosmoura , @drhaliburton 
I hope this helped.

Issue #1996, opened 2019-01-20
Blurred text content in md-dialog. Fonts? Wrong align?


MairwunNx:
![image](https://user-images.githubusercontent.com/39625750/51440300-b793b300-1cd6-11e9-9e54-e2fec7e7be17.png)

It not cool effect, but and it no fully bug, i think it can be caused wrong vertically align dialog, i get some effect when i learning css and try center div content when use translate with transform. 

I also see it effect with Roboto fonts. And also it can be see on vue-material site.

> ###### Steps to reproduce: it can be reproduce with random resizing window.

> ###### Which browser: Edge - Not blurred text. Google - Blurred text.

### Environment information
> ###### Vue-Material version [ `Latest` ]
> ###### System version: [ `Windows 10 x64` ]
> ###### Electron Api version: [ `4.0.0` ]
> ###### Chromium version: [ `69.0.3497.106` ]
> ###### Vue version: [ `2.5.21` ]

antoniotrkdz:
Why it was closed? This issue is still relevant.

LorhanSohaky:
any news?

PR #2023, opened 2019-02-22
fix(MdDialog):change position principle from transform to flex wrapper

hastom:
Currently dialog is positioned to screen center by appling styles:
```
left: 50%;
right: 50%;
transform: translate(-50%. -50%);
```
This is nice approach but it causes serios issues in Chrome (e.g. #1996). 
This happens because Chrome currently cannot properly render fonts when element has decimal pixels position. There are lots of attempts to fix it using hacks like `filter: blur(0.0001px)`, but the only stable solution, in pure css is to wrap element in container `display: flex; align-items:center; justify-content: center`. It works because flex-items always snap to integer-pixel position (demo https://codepen.io/hastom/pen/OdejXe).

This is a breaking change, because class now applies to wrapper-component and not directly to dialog. So if it will be approved, i think it should be included in v1 as it not released yet


marcosmoura:
This is a good fix. I will wait until next release, to test it a little bit more.  Thank you!

LorhanSohaky:
any news?

jbeacham:
Yeah this introduced a much worse bug in md-dialog. You can see it on the official website: [https://vuematerial.io/components/dialog/](https://vuematerial.io/components/dialog/) open any dialog.

PR #2024, opened 2019-02-23
in other to customise the icon

diadal:
in other to customise the icon


marcosmoura:
This can't be done like this. The user can opt-out to use the standard material library icon.

PR #2077, opened 2019-06-04
Finished MdTablePagination.vue component.

lukewarlow:
## Problem
This component is currently not finished, so is unusable.

This merge request fixes a few bugs within the current code, for instance the wrong event is used when setting page size, and finishes the implementation, specifically adding emitted events when the page is changed.

### DIFF
--- a/src/components/MdTable/MdTablePagination.vue
+++ b/src/components/MdTable/MdTablePagination.vue
@@ -61,7 +61,9 @@
         return ((this.currentPage - 1) * this.currentPageSize) + 1
       },
       currentPageCount () {
-        return this.currentPage * this.currentPageSize
+        let pageCount = this.currentPage * this.currentPageSize
+        if (pageCount > this.mdTotal) pageCount = this.mdTotal
+        return pageCount
       }
     },
     watch: {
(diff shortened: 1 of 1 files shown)


src/components/MdTable/MdTablePagination.vue
```diff
@@ -4,19 +4,19 @@
...
 
     <span>{{ currentItemCount }}-{{ currentPageCount }} {{ mdSeparator }} {{ mdTotal }}</span>
 
-    <md-button class="md-icon-button md-table-pagination-previous" @click="goToPrevious()" :disabled="mdPage === 1">
+    <md-button class="md-icon-button md-table-pagination-previous" @click="goToPrevious()" :disabled="currentPage === 1">
       <md-icon>keyboard_arrow_left</md-icon>
     </md-button>
 
-    <md-button class="md-icon-button md-table-pagination-next" @click="goToNext()">
+    <md-button class="md-icon-button md-table-pagination-next" @click="goToNext()" :disabled="currentPageCount === mdTotal">
```
bheptinh: `:disabled="currentPageCount === mdTotal"`
this is not correct, if I show 20 items per page and have total 35 items then on second page `currentPageCount` will be equal 40 and `mdTotal` - 35

lukewarlow: I will take a look at this and push a fix. thank you.

lukewarlow: I have just pushed a fix for this. Realised I had been passing through the wrong value into total so that I could never experience this bug.

PR #2079, opened 2019-06-06
Added on keypress enter event and onClear event to Autocomplete

fincha:
To make it possible to start async search on keypress Enter.


fincha:
can the test not fail?

>  SecurityError: localStorage is not available for opaque origins

hissalht:
> can the test not fail?
> 
> > SecurityError: localStorage is not available for opaque origins

The jest config does not have a "testURL" field. You just need to add it to the jest config in `package.json` [like so](https://github.com/vuematerial/vue-material/pull/2082/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R160)

fincha:
@marcosmoura can you halp me with the failing pipeline? I am basicly adding some event to autocomplete, this should be a problem with localstorage.

PR #2094, opened 2019-07-12
fix(MdMenu): check if element under event coords is inside menu

### LINKED ISSUE #2067: [mdAutocomplete] Not opening dropdown when clicking on label
### Steps to reproduce
- Use a **mobile phone** or chrome in **responsive mode**
- Go to [autocomplete example page](https://vuematerial.io/components/autocomplete/)
- Tap on the label _Country_ of the first autocomplete example
- The dropdown list will not appear

- If you tap the empty space in the input, next to the label, it will work (at the second attempt if the bug has been reproduced).

### Which browser?
Chrome 74 on **mobile** or **responsive mode**
Vue 2.5.2
vue-material 1.0.0-beta-11

### What is expected?
Open the autocomplete list even if I click the <label> element

### LINKED ISSUE #2093: [MdAutocomplete] menu closes when appears under cursor
Alt Text

This bug also affects any menus which open on MdInput focus

### Which browser?

Chrome 75

### What is expected?

Menu opens and stays open

Menu opens for a split second and closes

### Reproduction Link

https://vuematerial.io/components/autocomplete

hastom:
Resolves #2093
Probably resolves #2067

For some reason in Chrome document.click event triggers with target `<body>` when something appears under cursor during event loop.

Also method `isMenuContentEl` method is used to check if target is inside menu content. I think that's an expected behaviour cause it was approved in PR #1280

### DIFF
--- a/src/components/MdMenu/MdMenuContent.vue
+++ b/src/components/MdMenu/MdMenuContent.vue
@@ -1,5 +1,5 @@
 <template>
-  <md-popover :md-settings="popperSettings" :md-active="shouldRender">
+  <md-popover :md-settings="popperSettings" :md-active="shouldRender" @md-active="createObservers">
     <transition name="md-menu-content" :css="didMount" v-if="shouldRender" v-on="$listeners">
       <div
         :class="[menuClasses, mdContentClass, $mdActiveTheme]"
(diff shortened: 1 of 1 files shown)


hastom:
Should not be merged, closed in #2201

PR #2095, opened 2019-07-12
Inline version of md-datepicker

hastom:
New component. Core functionality copied from `MdDatepicker` but it provides much more flexible functionality to integrate it with other frameworks, standalone usage and to play with other vue-material components


hastom:
@marqbeniamin nice to see new labels on PR 😄 what information do you need?

PR #2106, opened 2019-08-05
Added Hide Icon prop

yshadmehr:


yshadmehr:
Add mdHideIcon to datepicker

PR #2157, opened 2019-12-05
observe tooltip content for changes

gaby64:
update position on tooltip change


gaby64:
useful for multi-purpose buttons, or cancelable single purpose buttons 
when direction is left

Issue #1679, opened 2018-04-12
Migrate from avoriaz to vue-test-utils


vjoao:
https://github.com/vuematerial/vue-material/blob/af0b89c5bbc6cbcf4bcce6bcdaf624bb54ed6192/test/utils/mountTemplate.js#L2

There is a deprecation note on avoriaz library: https://github.com/eddyerburgh/avoriaz#deprecation
We should move to vue-test-utils to maintain test compatibility.

Samuell1:
After changing to vue-test-utils mostly all tests for components are broken, because of changed names, functionality.

PR #2175, opened 2020-02-07
Migrate from avoriaz to vue-test-utils

mhio:
#1679

Still a WIP. 

- `mountTemplate` uses `createLovalVue()` to avoid test side effects, I've been bitten by a couple of obscure issues else where so I just avoid them.
- `.find('foo')[0]` to `.find()`
- `.find('foo')` to `.findAll('foo')`
- `.hasAttribute()`/.getAttribute() to .attributes()
- `.hasClass()` to `.classes()`
- Adding `await wrapper.vm.$nextTick()` for async updates. 
- Replacing `data` objects with functions. 

MdPortal needs some changes, @vue/test-utils doesn't seem to be able to mount MdPortal without a template.

### DIFF
--- a/src/components/MdPortal/MdPortal.test.js
+++ b/src/components/MdPortal/MdPortal.test.js
@@ -1,4 +1,4 @@
-import { mount } from 'avoriaz'
+import { mount, createLocalVue } from '@vue/test-utils'
 import mountTemplate from 'test/utils/mountTemplate'
 import MdPortal from './MdPortal'
 
@@ -8,6 +8,7 @@ test('should render the portal element inside body', async () => {
   const portalEl = wrapper.vm.$el
 
   expect(document.body.childNodes).toContain(portalEl)
+  wrapper.destroy()
 })
 
 test('should remove the portal element from destination before destroy', async () => {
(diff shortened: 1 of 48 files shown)


src/components/MdPortal/MdPortal.test.js
```diff
@@ -32,13 +33,16 @@ test('should render on a custom target', async () => {
...
   `
   await mountTemplate(MdPortal, targetTemplate, {
-    attachToDocument: true
+    attachToDocument: true,
   })
+ 
+  // debugger
+
   const targetEl = document.querySelector('.target')
   const portalWrapper = mount(MdPortal, {
```
mhio: Fails with `TypeError: Cannot read property 'child' of undefined` in https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/create-wrapper.js#L12

test/utils/mountTemplate.js
```diff
@@ -1,16 +1,21 @@
-import Vue from 'vue'
-import { mount } from 'avoriaz'
+import { createLocalVue, mount } from '@vue/test-utils'
 
 export default async (component, template, options = {}) => {
-  const newComponent = Vue.component(`${component.name}-test`, {
+  const localVue = options.localVue || createLocalVue()
```
mhio: This isn't necessary, but might avoid some head scratching.

PR #2181, opened 2020-02-27
[core] RTL support #80

tomdor123:
added rtl support to md-tabs component with mdIsRtl prop.
when in RTL the tab was moving to the wrong direction with the "transform: translate3D" property.
we created a hard coded fix in our project and it was making our life miserable, it would be awesome if this feature could be added.


dragosct10:
Hi, @tomdor123! Thanks for your work!

PR #2202, opened 2020-04-23
Eslint rules to force existing code style

hastom:
Just two more rules:
1. To force space before function paren's
2. To force space before closing tag

Most of the code is already matches them, so i think we should force them to prevent autoformatting issues for future contributors


hastom:
And one more to prevent semicolons :)

PR #2269, opened 2020-08-26
Fix typo in docs - Update Card.vue

lvcasjm:
hey, i spotted a typo while reading through the docs, heres a pr with what i assume is the correct wording.

I've never done a pr before so let me know if i did something wrong.

Cheers


marqbeniamin:
Hi @lvcasjm, great work! Thank you for PR, all is right! If you see other problems, we are waiting for PR! 🎉

PR #2108, opened 2019-08-10
Fixing the label for inputs when we received an invalid value

Topinaqui:
This fix resolves the following issue:
[ Issues with label and md-input type="Number" #2092](https://github.com/vuematerial/vue-material/issues/2092#issue-464762693)

Issue #2285, opened 2020-10-28
MdSelect "TypeError: Cannot read property 'badInput' of undefined"


cvalues:
This error occurs when an option is selected in the select:

Firefox:
TypeError: this.$el.validity is undefined

https://stackoverflow.com/questions/63994411/vue-js-error-in-nexttick-typeerror-cannot-read-property-badinput-of-undef

cvalues:
https://github.com/vuematerial/vue-material/issues/2270

seems like the same error

mullerivan:
same error here...

stephsalou:
same error ... 
![Kazam_screenshot_00101](https://user-images.githubusercontent.com/49046464/98230894-9c4d7280-1f53-11eb-9679-bb4efca0db65.png)

johnkristijan:
Hi,
I have created a pull-request for this one now.
[PR #2287](https://github.com/vuematerial/vue-material/pull/2287)

@marqbeniamin could you review it?

DevVue:
Yep same error, please merge if all good.

DJDeWitt:
Also getting this error as of today!

hansvn:
> Hi,
> I have created a pull-request for this one now.
> [PR #2287](https://github.com/vuematerial/vue-material/pull/2287)
> 
> @marqbeniamin could you review it?

While your PR fixes possible data integrity errors, it has been merged into the dev branch by now (#2262).
The change made there does not offer a solution to the root problem though:

When a value is selected in `MdSelect` it tries to get the DOM elements (`$el`) `validy` property.
While this works for `MdInput`, which has an `<input>` element as its root DOM element; `MdSelect`s `$el` is an `<md-menu>` which doesn't have the `validity` property.

A solution for this would be to add a reference to the correct `<select>` element.
I created a PR to fix the original intended behavior and provide a fix for this issue: #2288

leonardoks16:
Sorry. Can you make an example of using MdSelect with your fix?

leonardoks16:
Can you update this to npm package?

jkalousek:
I did overwrite the ...\node_modules\vue-material\ with current github dev version but i still get 

```
  [10364:1210/183001.221:INFO:CONSOLE(629)] "[Vue warn]: Error in nextTick: "TypeError: Cannot read property 'badInput' of undefined"

  found in

  ---> <MdSelect> at src/components/MdField/MdSelect/MdSelect.vue
[...]

PR #2287, opened 2020-11-06
Fixed badInput error message for MdSelect (#2285)

johnkristijan:
The MdSelect component is using a mixin (`MdFieldMixin.js`) that was causing a badInput error, even though this badInput validation was intended to fix #2108, it caused some unwanted side-effects, like this error.

In this commit I just added a check to see if the validation property exists.


marqbeniamin (approved):

ReseTTim:
Is this fix pushed to master?

PR #2288, opened 2020-11-12
MdSelect - Add reference to correct DOM element

hansvn:
Bugfix for #2285:
When a value is selected in `MdSelect` it tries to get the DOM elements (`$el`) `validy` property.
While this works for `MdInput`, which has an `<input>` element as its root DOM element; `MdSelect`s `$el` is a `<md-menu>` which doesn't have the `validity` property.

This bugfix adds a reference to the correct DOM element when checking for the select input validity.
It overrides the mixins `isInvalidValue` method with a reference to the correct DOM `<select>` element (which has the required `validity` property defined).

### DIFF
--- a/src/components/MdField/MdSelect/MdSelect.vue
+++ b/src/components/MdField/MdSelect/MdSelect.vue
@@ -42,7 +42,7 @@
     </div>
 
     <input class="md-input-fake" v-model="model" :disabled="disabled" readonly tabindex="-1" />
-    <select readonly v-model="model" v-bind="attributes" tabindex="-1"></select>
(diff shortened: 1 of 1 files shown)


Tropicalista:
Please can you merge this?

mrkoshi:
How long we will wait for this merge? Is there some problems with it?

Tropicalista:
@mrkoshi I think in dev branch is merged. However this library seems abandoned...

OladetounJed:
@hansvn , when will the NPM package be updated?

We are still having similar issues with the package

PR #2298, opened 2021-01-04
Fix/md tabs/ordering

### LINKED ISSUE #2015: md-tab not updated when data changes
I have a table of data that contains names and categories. 

I retrieve the unique category names and then use it in an `md-tab` set, for filtering the table result according to the selected group. But the tab labels behave strangely when the data is updated.

Specifically, some labels remain there but they should not. After data is updated, the tab indicators do not behave as expected either.

### Which browser?

Tested on both Chrome and Firefox

### What is expected?

- The tab labels are expected to be reactive on changes to "active groups"
- Tab indicators should disapear if the active item is removed.

- The tab labels do not reflect changes to the "activeGroups" data
- The tab indicator remains visible under an empty tab, taking up empty space as the tab has been removed on change to "activeGroups".

In comparison, using an `md-select` works. Console log also seems good.

### Reproduction Link

https://codesandbox.io/s/7jy1wnv410

### LINKED ISSUE #2282: [BUG] MdTabs: Tabs with number-like id are wrongly ordered by id
### Steps to reproduce

When md-tabs are Numbers or Strings having a Number value, tabs are ordered by id instead of by appearance order in the source code.

### Which browser?

Chrome on Windows.

### What is expected?

We declare tabs in a certain order in the Vue template: we expect to display the tabs in the same order.

Tabs are sorted by Number ids before displaying them.

### Reproduction Link

https://codesandbox.io/embed/vue-material-quick-start-forked-buewe

Here is the relevant part:

```vue
<template>
  <md-tabs>
    <md-tab :id="1" md-label="A">A</md-tab>
    <md-tab :id="2" md-label="B">B</md-tab>
    <md-tab :id="0" md-label="C">C</md-tab>
  </md-tabs>
</template>
```

Tabs do not appear in the order "A B C", but in the order "C A B".

### Possible Solution

I *think* this is due to the implementation of storing tabs in an Object and using its keys:

https://github.com/vuematerial/vue-material/blob/dev/src/components/MdTabs/MdTabs.vue#L81

```javascript
      MdTabs: {
        items: {}
      },
```

I *think* you could resolve this by using a new Map() instead of a raw {}.

I base my assumption on this test:

```javascript
a = {}; a["c"] = ""; a["a"] = ""; a["b"] = ""; Object.keys(a);
// Outputs ["c", "a", "b"]

a = {}; a[1] = ""; a[2] = ""; a[0] = ""; Object.keys(a);
// Outputs ["0", "1", "2"]

a = {}; a["1"] = ""; a["2"] = ""; a["0"] = ""; Object.keys(a);
// Outputs ["0", "1", "2"]

a = new Map(); a.set(1, ""); a.set(2, ""); a.set(0, ""); a.keys();
// Outputs MapIterator {1, 2, 0}
```

slaout:
This merge request fixes #2282 (and possibly also fixes #2015).

I heavily rebased to have one (small, meaningful and easy-to-review) commit per fix/feat.
Tell me what you think about it all.

There is a remaining WIP commit.
That commit is for you to visually test the behaviors before (by checking out the commit) and after (by checking out the branch) the bug fixes (same goes for unit tests before and after).
This big example lists all the possible edge case I could find, in order to do my best to not introduce a regression in the library.
I will remove that commit before merging.
Or if you have a private testbed where I could put the example's content, it would be even cooler: it is so easier to visually test weird edge-cases are correctly handled...
Unit tests are good for that too, but it's always nice to see it visually (and I didn't succeed to unit-test indicatorStyles generation)

About the implementation of ordered tabs (and numeric IDs): basically, there is no way for MdTabs or MdTab to know the order of tabs in the MdTabs' slot.
The only way I found is querying the DOM.
So I made MdTabs adds an ID object to the DOM object (and not an attribute, so the Number/String/Other nature of the ID is kept, and not transformed to a String when converted to a DOM-node attribute).
From this, I can compute an orderedIds array and base all order-sensitive computations on it (rendering, activation by index, swipe...).

There was a few bugs I found and fixed:
* activeTab is an ID, but it sometimes contained an index, and a double-bug reverted the bug sometimes, but not always
* missing null check on a duplicate activeTabIndex set
* the selector `.md-tab:nth-child(${this.activeTabIndex + 1})` did match sub-md-tabs too, and depending on the order of elements, a sub-tab's height could be misued as the height og the current tab
* md-changed is emitted twice when md-active-tab changes

### DIFF
--- a/docs/app/pages/Components/Tabs/Tabs.vue
+++ b/docs/app/pages/Components/Tabs/Tabs.vue
@@ -8,7 +8,7 @@
   <page-container centered :title="$t('pages.tabs.title')">
     <div class="page-container-section">
       <p>Tabs make it easy to explore, switch between different views and enable content organization at a high level, such as different data sets or functional aspects of an app.</p>
-      <p>Tabs are really powerfull and have deep integration with Vue Core features and router.</p>
+      <p>Tabs are really powerful and have deep integration with Vue Core features and router.</p>
       <note-block>More features for tabs will be come in the next weeks, like pagination scroll and touch events. :)</note-block>
     </div>
 
@@ -120,7 +120,7 @@
             {
               name: 'md-sync-route',
               type: 'Boolean',
-              description: 'Syncs the table selection with the current route, matching by the single tab <code>to</code> prop.',
+              description: 'Syncs the tab selection with the current route, matching by the single tab <code>to</code> prop.',
               defaults: 'false'
             },
             {
--- a/docs/app/pages/Components/Tabs/Tabs.vue
+++ b/docs/app/pages/Components/Tabs/Tabs.vue
@@ -72,6 +72,8 @@
 <script>
   import examples from 'docs-mixins/docsExample'
 
+  const TAB_ID_TYPE = 'String|Number'
+
   export default {
     name: 'DocTabs',
     mixins: [examples],
@@ -107,7 +109,7 @@
           props: [
             {
               name: 'md-active-tab',
-              type: 'String|Number',
+              type: TAB_ID_TYPE,
               description: 'Set the current selected tab. Works by providing the id of the desired <code>md-tab</code>.',
               defaults: 'null'
             },
@@ -209,7 +211,7 @@
         props: [
           {
             name: 'id',
-            type: 'String',
+            type: TAB_ID_TYPE,
             description: 'The tab id. Used when changing the active tab dynamically',
             defaults: 'a random string'
           },
--- a/docs/app/pages/Components/Tabs/Tabs.vue
+++ b/docs/app/pages/Components/Tabs/Tabs.vue
@@ -3,6 +3,7 @@
 <example src="./examples/TabContent.vue" />
 <example src="./examples/TabIcons.vue" />
 <example src="./examples/TabCustomTemplate.vue" />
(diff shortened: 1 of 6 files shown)


marqbeniamin:
Hi @slaout thank you for your PR. Sorry for the late response. I like your solution. Can you come back with an update on docs content? On the documentation page, you should put the "Tabs ordering" section and update the props table.

After your update, we come back with a new review and approve the PR.

All the best,
Ben

slaout:
Hi, thanks for your reply: no problem.

I'll look at your suggestions. But to be sure we're on the same page:

By 'put the "Tabs ordering" section', you mean I remove this testing section, not put it in a more permanent place?

And by 'update the props table' I tought I changed the type to "Number | String": what do you mean? (I have not the code under my eyes right now: it may or may not be obvious when I will read it).

slaout:
Done:

* I've rebased without the WIP commit showcasing the bugs before and after the fixes
* I've added a commit to document the new behavious
* As I was not sure about what you meant, I've added yet another commit to show a more user-friendly example of how tabs ordering work with dynamic tabs (can be squashed with the previous commit if it's OK with you, or removed if you did not mean to document this behaviour).

Sorry for the delay.

Looking forward for the new review :-)

PR #2314, opened 2021-03-05
fix(MdDatepicker): remove  in template

snarcraft:
Hello, 
I noticed that there is a forgotten `this`  in the template of the `MdDatepicker` component on conditionally adding `md-native` class


marqbeniamin:
@snarcraft thank you for your help 🙏

PR #2354, opened 2022-02-22
Fix: Add placement to datepicker dialog

### LINKED ISSUE #2353: MdDatepicker needs placement as a prop
### Why?

In some screen sizes if you put a date field at the right of the screen the md-datepicker-dialog opens out of the screen and adds a vertical scroll.

### How to fix it?
If we add an md-placement to the date-picker component we can place the dialog based on the input placement so in the case of input at the end of the screen we can set the placement to 'bottom-end' so popper js will place the dialog from the end.

### Other issues related to this issue?

https://github.com/vuematerial/vue-material/issues/2352
this issue also fixes this problem by setting the placement to 'bottom-end'.

payamnaghdy:
This MR closes #2353

### DIFF
--- a/src/components/MdDatepicker/MdDatepicker.vue
+++ b/src/components/MdDatepicker/MdDatepicker.vue
@@ -12,6 +12,7 @@
         :md-disabled-dates="mdDisabledDates"
         :mdImmediately="mdImmediately"
         @md-closed="toggleDialog"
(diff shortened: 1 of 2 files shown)


rory-wagner (approved):

marqbeniamin:
@payamnaghdy @rory-wagner Thank you for your help! 🙏

PR #2368, opened 2022-08-28
Fix: MdDatepicker scroll popover update

AriesAlex:
after reopening dialog window appears out of screen bounds
in the usual case mounted hook in mdpopover should fixing this, but here we using keep-alive in MdDatepicker so it's not

https://user-images.githubusercontent.com/39626061/187054220-c4d1f8da-bb3b-4fd7-b33c-5ce9d9b0bd9c.mp4

this PR automatically updates popper on activated hook (Called when the component is activated inside <keep-alive>)


AriesAlex:
" (Called when the component is activated inside )"
inside keep-alive
seems like it's not really like angle brackets

PR #2373, opened 2022-10-12
Feature: Adds prop to preserve dialog in the dom

### LINKED ISSUE #2346: Let Dialog get hidden [Feature Request]
On a page I'm working on, there is a button which opens up a Dialog (md-dialog) and contains a Progress Bar (md-progress-bar) with a stopwatch. The page, the Dialog and the Progress Bar reside in separate Vue components. As long as the Dialog is open the stopwatch runs fine. Now, I would like to show the time on the page, but when I click in the area outside of the Dialog, the Dialog closes and the time stops updating. The reason appears to be that, the Dialog is removed from the DOM and as I have seen there is no option to simply hide it in a v-show fashion. As I see here:
[ vue-material/src/components/MdDialog/MdDialog.vue ](https://github.com/vuematerial/vue-material/blob/4de86085fd9ef835df2caf534249bc7f6c21f5f4/src/components/MdDialog/MdDialog.vue)

"md-dialog" is implemented with a v-if directive:

`<div class="md-dialog" v-if="mdActive">`

I would like to ask for a feature request for making the Dialog able to get hidden and not completely removed. Maybe there is something I'm missing, so I'm also open to any suggestions.

franciscoemanuel:
Resolves #2346 

- Adds prop MdKeepAlive for MdDialog component that allows the render to happen using v-show;
- Adds prop documentation for the Dialog page;
- Adds test scenarios on MdDialog component for use cases of this new prop;
- Creates a new abstract component (MdDialogRender) that implements different render methods (v-show or a v-if like) to MdDialog component; 

> The use for window.dispatchEvent(new Event('resize')) was an alternative for elements that uses resize observers such as MdTabs to be rendered correctly inside the dialog


franciscoemanuel:
@marqbeniamin

marqbeniamin:
Thank you, @franciscoemanuel, for your help. I will merge this to the dev branch

